PR #4, opened 2019-04-11
Feature/custom layer manager

weberjavi:
So... this PR turned out to be bigger than it was intended 😅 
This are the things that are added here:
-  A couple of `thunks` (async API calls) have been added to the `routes` to fetch `featuredGlobe`and `dataGlobe` Web Scenes specs. This is used to be able to display the buttons to toggle the available layers (on `globe-featured-view-component`).
-  This `thunk` functions are lazy loaded to avoid the registration of the reducers on the store unless the route that triggers the thunk is joint.
-  A rudimentary `LayerManager` component has been created to toggle the scene layers. It gets updated every time the URL param `activeLayers` changes.
-  I've added some `selectors`, `utils` and `redux-modules` folders.
-  Tried [this](https://medium.com/@ktruong008/absolute-imports-with-create-react-app-4338fbca7e3d) but failed so added a couple of `alias` on the config file.

Feel free to open the discussion if anything does not seems ok 🙏 

PS: i've removed globe basemap from argis online (feel free to add someone if you like) 🌐 
Kapture 2019-04-11 at 19 25 45


aabdaab:
I added some small improvements in the code here, nothing that changes functionality. I moved registering modules to the main files of redux-modules so it's not done in the thunk itself

PR #5, opened 2019-04-12
Create location widget and add it to the scene

weberjavi:
This Pr adds a location widget to the map
[Pivotal](https://www.pivotaltracker.com/story/show/164925275)

It exposes arcgis [Locate widget](https://developers.arcgis.com/javascript/latest/api-reference/esri-widgets-Locate.html) as a component.
It is then used as `child` of the `webscene` component having then access to `view` and `map` props.

I think this is a nice way of explicitly  using arcgis ui [widgets](https://developers.arcgis.com/javascript/latest/api-reference/esri-widgets-Widget.html). Maybe we should move all the widgets we use on the views to be consumed like this, instead of setting some of them on the `sceneConfig` (that way we will be more explicit on what things we are displaying on the map)

Currently we are doing something like this:

`#featuredGlobeViewConfig.js`
```js
export default {
  environment: {
    atmosphereEnabled: false
  },
  ui: {
    components: ['zoom', 'navigation-toggle']
  },
  center: [-0, 0],
  zoom: 3
}
```
@aabdaab what do you think?

### DIFF
--- a/src/components/globe-featured-view/globe-featured-view-component.jsx
+++ b/src/components/globe-featured-view/globe-featured-view-component.jsx
@@ -5,6 +5,7 @@ import styles from 'styles/themes/scene-theme.module.scss'
 import ArcgisLayerManager from 'components/shared/arcgis-layer-manager';
 import { layerManagerToggle } from 'utils/layer-manager-utils';
 import Sidebar from 'components/shared/sidebar';
+import LocationWidget from 'components/shared/widgets/location-widget';
 
 const GlobeComponent = ({ sceneConfig, isSidebarOpen, sceneLayers, updateFeaturedGlobeQueryParam, activeLayers, query }) => {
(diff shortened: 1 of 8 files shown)


aabdaab:
I had to change the `LocationWidget` a bit because once rebased with develop where we already toggle layers, the re-render was triggered which was causing adding multiple location widgets (one for each render). I used hooks to fix that 🎉 
Agree about extracting all the widgets to react components; I added that change in this PR.
I'm gonna leave it open so you can go through the changes, we can discuss it on Monday.

weberjavi:
Hey @aabdaab nice hooks :)
By the way, trying to get the syntax of the `useEffect` hook,  found [this article](https://daveceddia.com/useeffect-hook-examples/) with some use cases that was helpful.

weberjavi:
I would approve this PR, but I opened it and github does not let me do it :D

PR #6, opened 2019-04-22
Add legend to map

aabdaab:
This PR adds a dummy legend component. 
Legend component is imported from `vizzuality-components` and for now displays a fake dataset for presentation purposes.
In this PR I tried to configure svg-sprite-loader but had different issues with it so in the end I include the icons necessary for the legend like so:
```
import { Icons as VizzIcons } from 'vizzuality-components';
...
<VizzIcons />
``` 
image


src/assets/icons/chevron.svg
```diff
@@ -1,3 +1 @@
...
-</svg>
\ No newline at end of file
+<svg width="14" height="14" xmlns="http://www.w3.org/2000/svg"><path d="M9.914 7.95l4.243 4.242a1 1 0 1 1-1.414 1.415l-4.95-4.95a1 1 0 0 1 0-1.414l4.95-4.95a1 1 0 0 1 1.414 1.414L9.914 7.95zm-6 0l4.243 4.242a1 1 0 1 1-1.414 1.415l-4.95-4.95a1 1 0 0 1 0-1.414l4.95-4.95a1 1 0 1 1 1.414 1.414L3.914 7.95z"/></svg>
```
weberjavi: It is unclear if we are using this `svg`s for the icons on the legend or the `vizzuality-components` ones.
Maybe we can get rid of this until we are able to use them (to avoid forgetting it if we do not finally get to config svg sprite loader on webpack config )

PR #7, opened 2019-04-24
Feature/landscape mode camera

weberjavi:
This Pr adds the `landscape-view-component` to the Scene.
It takes care of:
-  Updating `landscapeView` URL param depending if the user zooms in or out beyond a given zoom threshold (`zoomLevelTrigger`)
-  Change camera `tilt` the first time `landscapeView` param changes.

PS: gonna handle grid related functionality in another PR

**zoom controls**

Kapture 2019-04-24 at 13 37 21

**scroll wheel zoom**

Kapture 2019-04-24 at 13 41 23

**locate controls**

Kapture 2019-04-24 at 13 22 19


aabdaab:
Sometimes I get this error when I'm zooming in/out very fast (swiping touchpad):
image

weberjavi:
So... I think I've avoided the crash (haven't been able to reproduce it again) by making the switch to landscape view happen only when the map is not updating. 
This has two visible effects (one desirable and other improvable). 
The desirable one is that `tilt` change will only be triggered when the actual `camera travel` comes to an end (that is visible when clicking the location button and will be visible with the search widget on small areas, post codes...).
The improvable one is that tilt update only happens once camera and tiles finished its previous jobs, so it feels kind of strange and out of time. This is something we should work on, but I think is good enough for this first iteration.

Also changed zoom watcher logic to avoid changing state on every zoom change (that improved performance quite a lot).
I let this PR open to address further comments, but I'll be branching out from it to start playing with the grid layer 🌐

src/components/shared/landscape-view-manager/landscape-view-manager-component.jsx
```diff
@@ -0,0 +1,50 @@
+import { useState, useEffect } from 'react';
+
+const LandscapeViewManager = ({ view, zoomLevelTrigger, onZoomChange, query, isLandscapeMode }) => {
+  const [ isInteracting, setInteracting ] = useState(false);
+  const [ isUpdating, setUpdating ] = useState(false);
+  const [ landscapeView, setLandscapeView ] = useState(false);
+
+  const isLandscapeViewOnEvent = (zoomValue, landscapeView) => zoomValue >= zoomLevelTrigger && landscapeView === false;
```
aabdaab: can be simplified to `zoomValue >= zoomLevelTrigger && !landscapeView`

src/components/shared/landscape-view-manager/landscape-view-manager-component.jsx
```diff
@@ -0,0 +1,50 @@
...
+  const isLandscapeViewOffEvent = (zoomValue, landscapeView) => zoomValue < zoomLevelTrigger && landscapeView === true;
```
aabdaab: may be simplified to `isLandscapeViewOffEvent  = (zoomValue, landscapeView) =>  !isLandscapeViewOnEvent(zoomValue, landscapeView)`

aabdaab: also, maybe we can move these two functions to `selectors` file and pass it through props?

weberjavi: The proposed change to refactor `isLandscapeViewOffEvent` will not work. Not in all circumstances that `isLandscapeViewOnEvent` is `false` `isLandscapeViewOffEvent` is `true` (to trigger the change both, zoom and URL param, conditions should match)

src/components/shared/landscape-view-manager/landscape-view-manager-component.jsx
weberjavi: 👍

PR #9, opened 2019-05-08
Feature/globe component and pages

weberjavi:
So... in this PR several changes to the way the app is handling the state and its architecture have been made. Will try to explain the motivations and the way forward.
This decisions are not set in stone so are open to modifications, enhancements and what not. Be benevolent, this is made with the best of intentions :)

#### Architecture:
The changes made to the folder structure try to reflect better the actual app functionality and design.
-  A general layout for the app has been set (inside `layout/app-layout`), this will render pages depending on the route (through a `universal` component).
-  Featured and dataGlobe components have been substituted by pages and a general `Globe` component has been created to be used inside this pages.
-  `shared` folder has been removed and all components are on the same level (maybe `landscapeManager` and `layerManager` could sit inside a components folder on `Globe` component and widgets and other ui elements be inside a `ui` folder?)

#### State management:
After deep diving on Global Forest Watch approach to handle app state some improvements have been made to the way Half Earth manages state. Thanks @edbrett for the GFW architecture tour 🚴 
 on the last `I love your front end` meeting :>)
The things aren't changing a lot, we will still have some data inside dedicated redux store keys (mainly raw data needed to fill the contents of the different ui elements, charts...) and we will still have url params that will be used to define the actual state of the application (this is made with the help of `redux-first-router` that has its own `location` key in the store).
So, the things we've _stolen_ from GFW:
-  Have separated `keys` inside location to reference different types of app state. I've created the `globe` key to store app state related to the globe (currently are `activeLayers` and `landscapeMode` but eventually we could store there actual `zoom` level, globe `extent`, camera `centre` coordinates...). I've also set a `ui` key to store other states that do not directly reference globe internals but have influence on layout and widgets state (currently there's only `isPaddingActive`, it activates the sidebar and eventually will set globe container or globe scene padding). 
-  We've shamelessly stolen the `utils/stateToUrl` folder. It has some utilities to `parse` and `stringify` `query` params (doing some `atob` encoding, and also have the `setComponentStateToUrl` function that gets used inside components actions to update url state).
- So inside the pages selectors we merge the url state with the initial state defined for each page to get the actual state of the app and pass it to page components to define the way they should be displayed (both `globe` and `ui` state).

That approach help us to have a smaller store not losing the chance to have shared state among components to drive the interface. It also makes app layout state fully shareable.


weberjavi:
Sure @aabdaab, will fix conflicts and add some info to the README. 
Will let you know when it is ready to review it again.
Thanks!!

weberjavi:
@aabdaab ☝️ ready again with updated readme and fixed conflicts.

aabdaab:
Thanks! Gonna merge this so we can branch off of develop

PR #11, opened 2019-05-15
Feature/landscapeview fixes

weberjavi:
This PR fixes some quirks of the previous version of landscape view.
-  Updating camera `heading` when changing view.
-  Scale down view `extent` to avoid the painted grid to be cut off by the viewport edges.
-  Select the grid cell in the centre of the view when no aggregated grid cells are selected.
-  Change grid colours.

A `grid-dell-data` redux module is created to store the data of the selected grid cells to eventually be used on landscape widgets.


kfilip94:
@weberjavi Do we should test it only on data globe? If so, the data globe view doesn't work for me; `The ArcGIS API failed to load.`

weberjavi:
Yeah @kfilip94 it should be tested on the `dataGlobe`.
The API error was triggered when trying to load two layers that I've deleted now from the scene. It was not triggered always though, so try to reload or to test it in a minute time.
It should be working fine (once you enter your arcgis online account credentials)

aabdaab:
Hey, @weberjavi I tried to test it right now and it's not working for me either, I get the same `API` error

kfilip94:
It's a normal behaviour or a bug?
Screenshot from 2019-05-16 17 53 02
Screenshot from 2019-05-16 17 54 09
Screenshot from 2019-05-16 17 53 19

kfilip94:
Two issues below, I'll take a closer look tomorrow :)
- When I disable a layer, the selected grid stays in this same place:
jolno-72bqu


- Each time when we are moving the camera, without any layer enabled:

Screenshot from 2019-05-16 17 59 30

weberjavi:
Hey @kfilip94 about the last two issues, they are the expected behaviour. I'll try yo explain whats happening under the hoods.
We are using the `rarirty-richness-GRID` layer to get the geometries of the grid cells (it is activated by default on the globe `settings`). It opacity is set to zero so the user can not see it but we can access to it inside the `layerView` of the globe and perform the calculations to get the geometries of the contained grid cells and add the new (aggregated) geometry on the fly.
If we remove the layer from the globe the `landscape-manager` does not have access to the geometries of the cells, hence not being able to calculate or paint anything.
It happens to be this layer toggle on the sidebar but on the final version of the site there will be no way of toggling this grid layer, so it will be always available :)

weberjavi:
About the lines inside the aggregated grid cells, yeah it seems a bug :bug: 
Will need to investigate the `union` function from the `geometryEngine` to check what's happening. If you agree I might do that in another PR

kfilip94:
@weberjavi Thank you for the explanation, it makes sense ofc :+1:

weberjavi:
☝️ conflicts solved, ready to review

PR #12, opened 2019-05-16
Feature/esri widgets theme

kfilip94:
PIVOTAL: [ Zoom and location as URL params](https://www.pivotaltracker.com/story/show/166036505) | [Common theme of Esri widgets](https://www.pivotaltracker.com/story/show/165937404)
This PR:

- Sets our custom styles of Esri widgets,
- Adds `toggle-ui` component, now it does nothing but it'll be a starting point for implementing fullscreen mode,
- Sets location and zoom as query params, after refresh they are still present in the URL
84j44-1lriz


src/components/widgets/location-widget/location-widget.js
```diff
@@ -1,3 +1,54 @@
...
+  useEffect(() => {
+    loadModules(["esri/core/watchUtils"]).then(([watchUtils]) => {
+      setWatchUtils(watchUtils);
+    })
+  }, []);
+
+  // Update location in URL
+  useEffect(() => {
+    const watchHandle = watchUtils && locationWidget && watchUtils.whenTrue(locationWidget.view, "stationary", function() {
+      const { longitude, latitude } = locationWidget.view.center;
```
weberjavi: As you're defining `locationWidget.view` on line `20` as the `view` (passed as a `prop` from the `scene`), maybe this could be simplified and use `vie` instead of `locationWidget.view`:
```
const watchHandle = watchUtils && locationWidget && watchUtils.whenTrue(view, "stationary", function() {
      const { longitude, latitude } = view.center;
     handleLocationChange([longitude, latitude]);
    });
    return function cleanUp() {
      watchHandle && watchHandle.remove()
    }
  }, [watchUtils, view]);

  return null;
}
```

src/components/widgets/location-widget/location-widget.module.scss
```diff
@@ -0,0 +1,65 @@
...
+  -webkit-animation-iteration-count: infinite;
+  -webkit-animation-timing-function: linear;
+  -moz-animation-name: spin;
+  -moz-animation-duration: $durationTime;
+  -moz-animation-iteration-count: infinite;
+  -moz-animation-timing-function: linear;
+  -ms-animation-name: spin;
+  -ms-animation-duration: $durationTime;
+  -ms-animation-iteration-count: infinite;
+  -ms-animation-timing-function: linear;
```
aabdaab: I don't think these are needed, could you check?

src/pages/data-globe/data-globe-initial-state.js
```diff
@@ -1,6 +1,8 @@
 export default {
   globe: {
-    isPaddingActive: true
+    isPaddingActive: true,
+    zoom: 3,
+    center: [16.9515536, 51.116959]
```
aabdaab: why these values?

kfilip94: There should be some default values, latitude and longitude, we can change it to anything else.

src/pages/data-globe/data-globe-selectors.js
```diff
@@ -17,6 +18,14 @@ const getUiSettings = createSelector(selectUiUrlState, uiUrlState => {
   }
 })
 
+const getSceneSettings = createSelector(getGlobeSettings, globeSettings => {
+  return {
+    ...sceneSettings,
+    zoom: globeSettings.zoom,
```
weberjavi: Shouldn't this params belong to `globeSettings`? As they are params that change the way the `globe` appears?

kfilip94: As we agreed, I'll take care of handling state changes in the next PR.

kfilip94:
@weberjavi @aabdaab ready again :point_up:

PR #13, opened 2019-05-20
Get species categories & sidebar animations

aabdaab:
First iteration of this PT task: https://www.pivotaltracker.com/story/show/165937676
Now that I see it's tacking too many things for one PR, I decided to split it into couple of PRs for clarity, so you won't find any sidebar content in here :)
You will find though:
- Pulled out species categories from `rarity-richness-GRID` layer and put them into redux. They're categorized on reducer level (into **marine** and **terrestrial**). I added `side-effects` folder that resides on the `components` level for all sorts of side effects that are not introducing any direct ui change. I put there `BiodiversityDataManager`that needs `map` prop from `WebScene` so it's added as a child on the page level to the globe:
```
<Globe sceneId={SCENE_ID} sceneSettings={sceneSettings} >
    ...  
    <BiodiversityDataManager />
    ...
</Globe>
```
- Added `animations` key to url to handle, well, animations 😁 Using `onRest` property from react-spring to know when the entry boxes animation has ended so I can roll out sidebar. As the animation happens also on the initial load 
(1) I'm setting the `categoryBoxesAnimationEnded` key only when the category has been selected. This has some caveats as then (2) we need to remember to reset the animation flag on closing the sidebar so we can repeat point (1) - set it to false so that next time when we select a category have this effect working again. Not setting any animation flag for the sidebar as the entry boxes have some `delay` (100 - check in `entry-boxes-component`) and their closing/opening don't overlap.


src/side-effects/biodiversity-data-manager/biodiversity-data-manager-component.jsx
```diff
@@ -0,0 +1,26 @@
+import { loadModules } from '@esri/react-arcgis';
+import { useEffect } from 'react';
+
+const RICHNESS_RARITY_GRID = 'rarity-richness-GRID';
+const TAXA_FIELD = 'TAXA';
+
+const BiodiversityDataManager = ({ map, setSpecies, setSpeciesLoading, setSpeciesError }) => {
```
weberjavi: Maybe I'm kind of mistaken but this component looks like a `provider` to me. Wouldn't it make sense to it to sit on a `providers` folder. I mean, when I read `provider` I think of a component that feeds some data into the store.
Not sure if `side-effects` has a much broader meaning. What do you think?
Or is just me 🤷‍♂

aabdaab: I was always thinking of providers as services that are fetching data asynchronously from the backend and in here this is not the case as we're taking data from a layer (no direct call to the API). That's my interpretation 🤷‍♀ We need a third pair of eyes maybe

kfilip94: To be honest, I wrote only a few providers in my life so I'm not an authority here, but I took a look at how we solved providers in climate-watch and in my opinion it is similar to the provider's concept from CW. As Marta said, we are not fetching data asynchronously from the backend, we are passing `map` data as a prop, but also we are fetching some data `uniqueValues` through 'loadModules' from ArcGIS API. So to me, it looks more like provider, we are fetching part of the data in an async action and then we are passing the data to the store. Correct me if I'm wrong
provideers

aabdaab:
Ok, I think this is again ready for review ☝️ 
I improved the animations and moved the BiodiversityManager code to the `onLoad` callback.
I also added radio buttons for switching layers (no handlers are yet hooked or anything, just styling).

aabdaab:
P.S. Don't get so attached to the details of radio group and radio buttons in terms of its styling, this is a first iteration, I'm aware of the missing padding, missing toggle etc. The purpose of this first touch was to do custom radio buttons and write on top of the native ones. Final design completion coming up in the next PR! stay tuned 👀

src/styles/settings.scss
```diff
@@ -3,6 +3,10 @@ $white: #fff;
 $elephant: #0F2B3B;
 $robinsEggBlue: #00BDB5;
 $osloGray: #838C93;
+$alto: #DBDBDB;
+$firefly: #09141A;
+$vida-loca: #4B9217;
```
kfilip94: Are we going to use camel case or snake case?

aabdaab: aah sorry. `camelCase` I think

weberjavi: Not really sure if I like this naming 🤔 
Maybe something to think about. From the name I can not say the colour it is. Maybe something more project related (like `primaryColor` or `titleGray`) or even colour related (`deepBlue`, `paleBlue`) could be easier to remember.
Nothing to do with this PR though, just opening the debate.

aabdaab: The thing is if you've got couple of shades of the same color you quickly run out of names for them (i.e light-blue, dark-blue, medium-blue, ...).
What I was doing in other projects if some of the colors were used more frequently than the others, then I'd give them a more semantic name like `theme-color` or `primary-blue` or sth similar. i.e.:
```
// all colors
$robinsEggBlue: #00BDB5;
$alto: ...
...

// frequently used
$widgets-color: $robinsEggBlue;
...
```
Would that help?

kfilip94: @aabdaab That's a nice approach, in the past, I also had problems with colors naming e.g. (grey, darkGrey, darkestGrey, darkAsHellGrey), the problem is that at some point this naming convention can be misleading either. But your solution above makes sense to me :+1:

PR #15, opened 2019-05-22
Feature/add search widget

kfilip94:
This PR adds the implementation of Search widget to the globe:
[PIVOTAL](https://www.pivotaltracker.com/story/show/165939533)

- adds styled SearchWidget,
- adds new folder `override` with custom styles of esri widgets, let me know if this makes sense to you,
Screenshot from 2019-05-22 12 46 20
- autocomplete in placeholder is not implemented, maybe I'll try in next PR,

a0a90-i1au0


src/components/widgets/search-widget/search-widget-components.jsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
```
weberjavi: Just a small typo on the component name, it should be `search-widget-component.jsx` in singular (to match the convention of the rest of the components).

src/components/widgets/search-widget/search-widget.js
```diff
@@ -0,0 +1,62 @@
...
+        popupEnabled: false, // hide location popup
+        resultGraphicEnabled: false // hide location pin
+      });
+      setSearchWidget(sWidget);
+
+    }).catch((err) => console.error(err));
+  };
+
+  const handleCloseSearch = () => {
+    view.ui.remove(searchWidget);
```
weberjavi: Maybe @faustoperez or @fanny can add something here but I think that they said that it would be great that the globe behind the full screen search started to move before the search disappears, maybe some timeout would do the trick?

src/components/widgets/search-widget/search-widget.js
```diff
@@ -0,0 +1,62 @@
...
+      searchWidget.viewModel.on("search-start", handleCloseSearch);
+      searchWidget.watch('activeSource', function(evt) {
+        evt.placeholder = "Search for a location";
+      });
+    }
+
+    const node = document.createElement("div");
+    view.ui.add(node, "top-left");
+    const component = <SearchWidgetComponent handleOpenSearch={handleOpenSearch} handleCloseSearch={handleCloseSearch} showCloseButton={!!searchWidget}/>;
+    ReactDOM.render(component, node);
```
weberjavi: Maybe I'm missing something of the internals of `esri widgets` customization, but why are we creating a DOM node and adding the component to that node instead of returning it at the end of `SearchWidget` function?
And place it on the screen using `scss` instead of the Esri `ui` positions.
Not saying it has to be like this, just curious.

src/styles/settings.scss
```diff
@@ -2,6 +2,7 @@
 $white: #fff;
 $elephant: #0F2B3B;
 $robinsEggBlue: #00BDB5;
+$osloGray: #838C93;
```
weberjavi: `$robinsEggBlue`, `$osloGray`...? 
You have a hidden poet inside @kfilip94 😅

kfilip94: Sadly it's not me :disappointed:, I use this site:
http://chir.ag/projects/name-that-color/#6195ED

src/components/widgets/search-widget/search-widget-components.jsx
kfilip94: Done

src/components/widgets/search-widget/search-widget.js
kfilip94: :man_facepalming: :woman_facepalming:  I forgot how to React... fixed

src/components/widgets/search-widget/search-widget.js
kfilip94: as @fanny said on slack: `we don't remember thinking about that globe movement before the search dissapears`

weberjavi: 👍 sorry :/

src/components/widgets/search-widget/search-widget.js
aabdaab: @kfilip94 I remember I asked you this before and you told me that it had the advantage of positioning it with esri's `top-left` that would always adjust to the panning and other components being rendered on this side of the screen. Not sure if it's still relevant.

kfilip94: Yeah, an advantage of using `view.ui.add` and `ReactDOM.render` is that position of your component will be placed according to other esri widgets, so in this case, you don't have to specify CSS like  `top` or `left` properties. However, the `Find Places` button is always on the top-left corner of the viewport so setting a position with CSS is very easy. Also, `SearchWidgetComponent` is more complicated, the `Find Places` button depends on the `search widget` side effects and local state, so we have to also take care of removing `Find Places` button from the DOM.

PR #17, opened 2019-05-29
Existing protection sidebar

aabdaab:
This PR adds a content of `Existing protection` category to sidebar with toggling WDPA data layer on a map:

image


src/pages/data-globe/data-globe-component.jsx
```diff
@@ -5,28 +5,28 @@ import LandscapeViewManager from 'components/landscape-view-manager';
...
+  { name: 'Urban pressures', value: 'urban' }
 ];
 
-const checkboxesFishingActivity = [
-  { name: 'Drifting longlines' },
-  { name: 'Fixed gear' },
-  { name: 'Other' },
-  { name: 'Purse seins' },
-  { name: 'Trawlers' }
+const checkboxesWDPA = [
```
weberjavi: Just thinking that maybe this should be in the constants folder, as is something that we may be using on the landscape widgets.
What do you think?

aabdaab: yes, definitely, I moved it 🙆

aabdaab:
rebased and merging, will deploy develop to `demo` url as well

PR #18, opened 2019-05-29
Feature/human presures widget

weberjavi:
This PR is the first iteration of the Human Pressures widget for landscape view.
[InVision](https://projects.invisionapp.com/d/main#/console/17484384/363015154/preview)

Some bits are missing in this version:
-  Info button.
-  Clickable treemap to activate the layers.
-  There's a bug with the tooltips (`react-tooltip`). The tooltip is only working as expected on the first render (and neither sometimes) :(

Kapture 2019-05-29 at 18 57 36


kfilip94:
@weberjavi Well done!
I think the color of cells is different in the design:
Screenshot from 2019-05-30 11 47 51

There are a few things but probably  for a different PR:

- I see an issue with cell size and text clipping:
Screenshot from 2019-05-30 11 41 58
- It'd be nice to add new constants to `settings.sccss` with color values
- Also console errors, but not now maybe :)

weberjavi:
Hey @kfilip94 about treemap cells styles:
there are two different states one when the layer is visible and another when it is not visible. The classes are there on the styles .scss file but we are no activating the layers yet so this styles are not being applied.

Also about text clipping, when there's no enough space to display the text a tooltip should be triggered (right now there's a [bug](https://www.pivotaltracker.com/story/show/166351701)). Maybe that's something that we could take a look on a design review on the next iteration (once we are able to activate human pressure layers)

image

I'll move the colours to the `settings` files 👍

weberjavi:
ready again ☝️

PR #23, opened 2019-06-03
Legend styling, modal + full screen mode

aabdaab:
This PR adds:
* legend **styling** (in the next one all handlers including handling opacity)
* **handling** full screen mode
* info buttons to sidebar with **modal opening** (no content there) 

PT tasks: 
https://www.pivotaltracker.com/story/show/166375072
https://www.pivotaltracker.com/story/show/166375230
Legend styling didn't have a PT task.

image

Also, note that legend still needs some styles tweaking, for example changing the _opacity_ icon although it might not be so straightforward due to possibly adding `svg-sprite-loader` to render custom icons which before was causing some problems...


aabdaab:
rebased with develop & ready to review

kfilip94:
1. Please change z-index of `aboutPage` class in `about-styles.module.scss` file to 3 or higher:
Screenshot from 2019-06-04 10 51 08
2. About the info buttons - shouldn't we show them on hover as well? And shouldn't they be aligned to the right side? (I can't really check it because InVision is down now :( )
Screenshot from 2019-06-04 10 58 21

src/components/legend/legend-styles.module.scss
```diff
@@ -1,37 +1,107 @@
+@import 'styles/settings';
```
kfilip94: I think this file needs refactoring, but I assume it'll happen anyway because this is the first iteration.

src/components/legend/legend-title/legend-title-component.jsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import styles from './legend-title-styles.module.scss';
+import molLogo from 'logos/mol_short_logo.png';
+
+// const legendName = (activeLayer, name) =>
```
kfilip94: Do we need this here?

src/components/modal/modal-component.jsx
```diff
@@ -0,0 +1,114 @@
...
+    <Modal
+      onRequestClose={handleClose}
+      isOpen={isOpen}
+      theme={styles}
+    >
+      <h2>{title}</h2>
+      {
+        loading ? <Loading height={200} /> : (
+          <>
+            {
```
kfilip94: Can we move `metadata` condition to the top somehow, without repeating it below?

src/components/sidebar/sidebar-component.jsx
```diff
@@ -8,11 +8,11 @@ import FixedHeader from 'components/fixed-header';
 import uiStyles from 'styles/ui.module';
 import styles from './sidebar-styles.module.scss';
 
-const Sidebar = ({ map, theme, children, activeCategory, handleSidebarToggle, isSidebarOpen, isLandscapeMode }) => {
+const Sidebar = ({ map, theme, children, activeCategory, handleSidebarToggle, isSidebarOpen, isLandscapeMode, isFullscreenActive }) => {
   const slide = useSpring({
     from: { marginLeft: -400 },
-    marginLeft: isSidebarOpen && !isLandscapeMode ? 0 : -400,
-    delay: isSidebarOpen && !isLandscapeMode ? 400 : 0
+    marginLeft: isSidebarOpen && !isLandscapeMode && !isFullscreenActive ? 0 : -400,
```
kfilip94: I think we can think about how to simplify this kind of condition chain, after the first release ofc :)

kfilip94:
@aabdaab Thanks, nice job! Check my comment above ;)

src/components/modal/modal-component.jsx
aabdaab: i will be changing this file anyway once we have metadata, i'll leave it like this for now

PR #27, opened 2019-06-05
Fix/biodiversity radio buttons

tsubik:
Fixing biodiversity layers sidebar.

Using activeLayers state instead of RadioButton inner component state which is probably destroyed when we close that part of layers manager.

fix-biodiversity-layers

[PIVOTAL](https://www.pivotaltracker.com/n/projects/2321903/stories/166441788)


tsubik:
I have no idea how to bring back cursor to my screen recorder, so sorry about that poor video.

PR #28, opened 2019-06-05
Hook up legend

aabdaab:
* Wire legend to display active layers
* Add legend configs for human impact and protected areas
* Dynamic title for Human Impact layer with 3 different rasters
* Handle opacity of layers through the legend
* Handle removing layer through the legend


kfilip94:
I found two bugs, I added them to PT, probably we will handle them later:
[BUG 1](https://www.pivotaltracker.com/story/show/166469901)
[BUG 2](https://www.pivotaltracker.com/story/show/166470514)

PR #31, opened 2019-06-10
Feature/describe landscape using geodescriber

tsubik:
image


tsubik:
Only thing is that it does not work for some areas. Geodescriber API must have some bug as I'm providing valid geojson and it is responding with 500.

image

PR #32, opened 2019-06-10
Feature/contentful metadata

weberjavi:
This PR adds contentful metadata to the store to fill the metadata modal of the different layers on the sidebar.
Kapture 2019-06-10 at 14 55 59


src/components/radio-group/radio-group-component.jsx
```diff
@@ -1,39 +1,23 @@
...
-  options.forEach(option => {
-    Object.keys(option.layers).forEach(variant => {
-      optionsLayers.push({
-        layerTitle: option.layers[variant],
-        variant,
-        option
-      });
-    });
-  });
+  const [variant, setVariant] = useState(RARITY);
```
tsubik: Hmm, that won't work. As the selected variant source of truth is the active layer that info should be taken from activeLayers. When this component is destroyed we will use that inner state. Try to close Biodiversity Layers and open them again. You will see that it is not selecting radio buttons correctly.

weberjavi:
@tsubik ☝️ ready again

src/components/checkbox-group/checkbox/checkbox.js
```diff
@@ -1,3 +1,25 @@
...
 import Component from './checkbox-component';
+import actions from 'redux_modules/metadata';
 
-export default Component;
\ No newline at end of file
+const CheckboxContainer = props => {
+
+  const handleInfoClick = (option) => {
+    const { setModalMetadata } = props;
+    console.log(option)
```
tsubik: forgotten console log

weberjavi: -.-

weberjavi:
ok, already removed.
so... merging

PR #36, opened 2019-06-12
Feature/use contentful to show texts

kfilip94:
This PR display texts from `contentful` in Half-Earth-modal title and text, also titles and descriptions in `About/Partners` page. In this PR I removed CartoDB logo from partners page as well.
[PIVOTAL - contentful](https://www.pivotaltracker.com/story/show/166604035) | [PIVOTAL - CartoDB logo](https://www.pivotaltracker.com/story/show/166604126)

**NOTE:** Before starting, run `yarn install`, I added `ReactMarkdown` package to the project


weberjavi:
reviewing ☝️

src/redux-modules/page-texts/page-texts-actions.js
```diff
@@ -0,0 +1,25 @@
...
+  const { pageTexts: { data }} = state();
+  if (isEmpty(data)) {
+    dispatch(fetchPageTextsData());
+  }
+});
+
+export const fetchPageTextsDataFail = createAction('fetchPageTextsDataFail');
+export const fetchPageTextsDataReady = createAction('fetchPageTextsDataReady');
+
+export const fetchPageTextsData = createThunkAction('fetchPageTextsData', () => async dispatch => {
```
weberjavi: The slug should be passed to the contentful service
```js
export const fetchPageTextsData = createThunkAction('fetchPageTextsData', slug => async dispatch => {
  try {
    const data = await CONTENTFUL.getTexts(slug);
    dispatch(fetchPageTextsDataReady({ data }));
  } catch (e) {
    console.warn(e);
    dispatch(fetchPageTextsDataFail(e));
  }
});
```

src/services/contentful.js
```diff
@@ -82,4 +89,12 @@ async function getStories() {
   return null;
 }
 
-export default { getEntries: fetchContentfulEntry, getMetadata, getStories };
\ No newline at end of file
+async function getTexts() {
```
weberjavi: I would do the same thing as we do with the metadata. Instead of fetching all the text entries at once would be preferable to fetch them once they're needed. Passing some filter params will do the trick:
```js
async function getTexts(slug) {
  const data = await fetchContentfulEntry({ contentType: 'texts', filterField:'view', filterValue: slug });
  if (data && data.items && data.items.length > 0) {
    return parseTexts(data.items);
  }
  return null;
}
```

src/components/widgets/minimap-widget/minimap-widget.js
```diff
@@ -1,8 +1,13 @@
 import { loadModules } from '@esri/react-arcgis';
 import React, { useState } from 'react';
+import { connect } from 'react-redux';
+
 import MinimapWidgetComponent from './minimap-widget-component';
 import { disableInteractions, minimapLayerStyles, synchronizeWebScenes } from 'utils/minimap-utils';
 import HalfEarthModal from 'components/half-earth-modal/half-earth-modal';
+import actions from 'redux_modules/page-texts/page-texts';
```
weberjavi: We do not need to import the `.js` file `'redux_modules/page-texts/page-texts'`, it will work with `'redux_modules/page-texts'`.
We are using the `DirectoryNamedWebpackPlugin` on the `config-overrides` so if the folder contains any file with the same name it will treat it as if it was the entry point :)

src/pages/data-globe/data-globe.js
```diff
@@ -5,9 +5,10 @@ import { layerManagerToggle, exclusiveLayersToggle, layerManagerVisibility, laye
 import Component from './data-globe-component.jsx';
 import mapStateToProps from './data-globe-selectors';
 import biodiversityActions from 'redux_modules/biodiversity-data/biodiversity-data';
+import pageTextsActions from 'redux_modules/page-texts/page-texts';
```
weberjavi: I think this is not needed here (as we are importing the action on the `connect`ed components `minimap-widget` and `about`)

kfilip94:
@weberjavi thank you for detailed feedback, it was super-helpful! :+1:  I fixed these things, it's ready again :racehorse:

PR #37, opened 2019-06-12
Add share modal

aabdaab:
This PR adds a share modal with a sharable link and an embedded one.

image

On the design there are `coordinates` and `orientation` but it looks like it was designed with Cesium stack in mind cause for example in Esri there are no `pitch` and `roll` values that would indicate the orientation. There is a `heading` however with which I was experimenting but it wasn't working as expected. It seems that longitude, latitude and zoom level are enough for Esri to persist a globe state. Test with changing a location, rotating etc, then copy the url and see that globe state is the same in a different tab.


simaob:
Hi Marta, about the coordinates, if we don't need them for ESRI, we can replace whatever is in that bit of text with the relevant params for the ESRI globe state.

aabdaab:
In this case `coordinates` are just longitude and latitude ☝️

kfilip94:
:arrow_up:  reviewing

aabdaab:
ready again ☝️

kfilip94:
hey @aabdaab, good job, below two things I found: 

- Facebook link doesn't work, does this occur because of localhost or just a bug?
Screenshot from 2019-06-12 17 47 40

- I think you can set a delay on `COPY` button. After clicking you can set a time interval, after 2 or 3 secs you can change the text to `COPY` again and make it clickable. I saw this once and I think it is a nice solution.

I'll take a look at the code :mag:

aabdaab:
1. good idea about the delay, i'll add it
2. facebook doesn't allow sharing localhost so this issue will disappear once we ship it 🚢

src/components/legend/legend-component.jsx
```diff
@@ -22,6 +22,37 @@ const HELegend = ({ map, datasets, handlers, isFullscreenActive, visibleLayers,
 
   const { layers } = map;
 
+  const handleStyle = {
+    border: '1px solid #0E2B3B',
```
kfilip94: Please use `$blue` (#0E2B3B) constant here

src/components/legend/legend-component.jsx
```diff
@@ -22,6 +22,37 @@ const HELegend = ({ map, datasets, handlers, isFullscreenActive, visibleLayers,
...
+    backgroundColor: '#FFFFFF',
+    height: '10px',
+    width: '10px',
+    boxShadow: '0 2px 4px 0 #0E2B3B',
```
kfilip94: And here

src/components/legend/legend-component.jsx
```diff
@@ -22,6 +22,37 @@ const HELegend = ({ map, datasets, handlers, isFullscreenActive, visibleLayers,
...
+    borderRadius: '50%',
+    position: 'absolute',
+    transform: 'translate(-5px,-9px)',
+    cursor: 'pointer',
+    outline: 'none'
+  };
+
+  const trackStyle = { 
+    backgroundColor: '#1bcec7',
```
kfilip94: Please use `$java` (#1bcec7) constant here

src/components/share-modal/share-modal.js
```diff
@@ -0,0 +1,28 @@
+import { connect } from 'react-redux';
+import get from 'lodash/get';
+
+import Component from './share-modal-component';
+
+const shareSocialMedia = [
+  {
+    link: `https://www.facebook.com/sharer/sharer.php?u=`,
+    className: 'facebookIcon',
```
kfilip94: Here, instead of className as a string, you can import directly class from styles file:
```code
    import styles from './share-modal-styles.module';
```
```suggestion
    className: styles.facebookIcon
```

src/components/share-modal/share-modal.js
```diff
@@ -0,0 +1,28 @@
...
+    alt: 'Facebook'
+  },
```
kfilip94: I think, you can do similar thing with icons:
```
import facebookIcon from 'icons/facebook.png';
import twitterIcon from 'icons/twitter.png';
```
```suggestion
  icon: facebookIcon
},
```

src/components/share-modal/share-modal-component.jsx
```diff
@@ -0,0 +1,110 @@
...
+          <Button
+            onClick={() => window.open(`${socialMedia.link}${currentLocation}`)}
+            theme={{
+              button: cx(styles.iconBackground, {
+                [styles.facebookIcon]: socialMedia.className === 'facebookIcon'
+              })
+            }}
+            key={socialMedia.className}
+          >
+          <img src={socialMedia.className === 'facebookIcon' ? facebookIcon : twitterIcon} alt={socialMedia.alt} />
```
kfilip94: After [this change](https://github.com/Vizzuality/half-earth-v3/pull/37/files#r293284621), you can do:
```suggestion
          <img src={socialMedia.icon} alt={socialMedia.alt} />
```

src/components/share-modal/share-modal-component.jsx
```diff
@@ -0,0 +1,110 @@
...
+      <div className={styles.coordinates}>
+        <p>coordinates: [{coordinates[0].toFixed(3)}, {coordinates[1].toFixed(3)}]</p>
+      </div>
+      <div className={styles.socialMediaContainer}>
+        {shareSocialMedia.map(socialMedia => (
+          <Button
+            onClick={() => window.open(`${socialMedia.link}${currentLocation}`)}
+            theme={{
+              button: cx(styles.iconBackground, {
+                [styles.facebookIcon]: socialMedia.className === 'facebookIcon'
```
kfilip94: After [this change](https://github.com/Vizzuality/half-earth-v3/pull/37/files#r293284621), you can pass class directly:
```suggestion
               socialMedia.className
```

kfilip94:
Hey @aabdaab, check my comments above :point_up_2:

aabdaab:
ready again ☝️

PR #38, opened 2019-06-12
Add arrow icons

aabdaab:
Add arrows to Biodiversity Layers in the sidebar:
iconsSwitch

### DIFF
--- a/src/assets/icons/switch.svg
+++ b/src/assets/icons/switch.svg
@@ -1,0 +1,1 @@
+<svg width="6" height="9" xmlns="http://www.w3.org/2000/svg"><path d="M5.392 4.438L.713.5v7.877z" fill="#00BDB5" fill-rule="nonzero"/></svg>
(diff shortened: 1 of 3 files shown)


kfilip94:
@aabdaab Great job, this switch looks much better with the arrow! :+1:
One thing that could be improved is a position of Info icon, I'd prefer to move this icon to the left side, next to the layer title. This would be more coherent with checkbox styles, what do you think?
(below gif with old and then with new position)
xkoni-ms4yk

aabdaab:
It does look better, true, but I think the idea is that the metadata is about rarity and richness rather than the taxa itself... So, visually it should belong to the toggle switch like it does now.

kfilip94:
So maybe this way? We can ask Fausto or Fanny about that.
Screenshot from 2019-06-12 15 15 42

fannycc:
Hiya!

aabdaab:
Fixed the arrows ☝️

PR #39, opened 2019-06-12
Implement active layers counters for entry boxes

kfilip94:
This PR adds counters to entry boxes:
[PIVOTAL](https://www.pivotaltracker.com/story/show/166617858)

Screenshot from 2019-06-12 12 51 13

### DIFF
--- a/src/components/entry-boxes/entry-boxes.js
+++ b/src/components/entry-boxes/entry-boxes.js
@@ -1,13 +1,12 @@
 import React from 'react';
 import { connect } from 'react-redux';
 import Component from './entry-boxes-component';
-
+import { getCountedActiveLayers } from './entry-boxes-selectors';
 import * as actions from 'actions/url-actions';
 
 const EntryBoxesContainer = props => {
   const openSidebar = () => props.changeUI({ isSidebarOpen: true });
-  const setActiveCategory = 
-    (activeCategory) => props.changeUI({ activeCategory: activeCategory });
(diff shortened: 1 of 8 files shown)


src/components/entry-boxes/entry-boxes.js
```diff
@@ -1,13 +1,12 @@
 import React from 'react';
 import { connect } from 'react-redux';
 import Component from './entry-boxes-component';
-
+import { getCountedActiveLayers } from './entry-boxes-selectors';
```
weberjavi: @aabdaab , @kfilip94 just opening this conversation.
I prefer the pattern we where doing lately of exporting `default` from selectors files an structured selector (`createStructuredSelector({ ...selectors map in here... }))`) an doing `import mapStateToProps from './selectors'` than importing the named selectors and creating the `mapStateToProps` on the entry file. I think the component container file it is cleaner and having the structured selector at the end of the selectors file also looks quite terse.
What do you think?

aabdaab: I also prefer this way. If you create a `selectors` file anyway you might just as well export the `createStructuredSelector` there; this way it's a lot easier to add more props to the component without lots of changes

kfilip94: I agree with you, it can be a cleaner approach. I'm going to change it now :+1:

kfilip94:
@weberjavi, @aabdaab ready again :point_up: :point_up_2:

PR #40, opened 2019-06-12
Feature/landscape human pressures toggle

weberjavi:
This PR allow users to toggle human pressure layers from landscape side bar widget.
I've updated selected layer styles on the treemap and the tiling mode of it. Also added some formatting to the values of the pressures and added a fallback statement to be displayed when there's no human pressures on the selected areas.
I didn't manage to get the tooltip working though :(
Kapture 2019-06-12 at 14 06 58

PS: I've been testing the new facets layer on the featured map. It should be working there but lets try if it works on staging before adding it to the dataGlobe


aabdaab:
The functionality's working great, I have some doubts about the styles, maybe we could improve them in this PR (or another? as you prefer)
1. Text sometimes is going of bounds; we should talk to designers what to do in cases like that, maybe use ellipsis and show the full title in the tooltip?
image
2. Also, there are cases when one of the human pressures rasters have very small percentage and become hard to see and click:
image
Not sure how to fix that if it should reflect the real percentage..

weberjavi:
@fannycc , @faustoperez any thoughts on that ☝️ ?

fannycc:
Not sure what's feasible here in terms of implementation. I was imagining that no matter the percentage is, the shape would remain square. 

- Also it looks like we are dividing the **squares vertically**, can we also dividing them horizontally? so It'd improve the spacing for texts... 
-Also it looks like there is an overall opacity. If you check the UI KIT all has solid colors and the active ones have the white border. "No pressure looks non existent there"
-Can we set a minimum so the smallest percentage is clickable?
-If the text doesn't fit there we don't need to put anything, the user will hover to find out what it is. So the cases would be 1.all the info fits / 2. Only the number of the percentage // 3. no text

I don't know if this can be helpful for you guys :/ let me know. If we don't find an easy solution we can ask around

weberjavi:
Hey @fannycc thanks for the comments.
Some thoughts:
- Regarding the square shape of the treemap there are a few [possible tiling options](https://bl.ocks.org/d3indepth/fa5e9d42d8e260f3f76a98be648c9edd) to render the chart, but none of them ensure a square shape (per definition a treemap recursively subdivides area into rectangles). Take a look to the options and tell me which one fits better. I've tested all of them and ended up using `binary`.
- We do not have a lot of control of the shape the chart takes, so if we wanted divided horizontally they will be always like that (and I think it looks worst, but check it, is the `slice` tiling option).
- About the opacity on the `No pressure` 'square' I've set the opacity to `0.5` as it is set on the inVision on the landscape screen and on the UI kit :P 
- I'm not sure setting a minimum width and height for layers with low values will be a good idea, but would be preferable if @greta or @benlaken jump in. For me this is a data visualisation that happens to be clickable and not _au contraire_
- Will try to tweak the logic for hiding the text, but I would only go with options 1 and 3 avoiding the ellipsis.

BTW this branch is available here:
https://half-earth-demo.vizzuality.com/dataGlobe

Feel free to take a look to check several layouts of the treemap on different grid cells, it changes a lot :)

aabdaab:
i.e. check this use case: https://half-earth-demo.vizzuality.com/dataGlobe?globe=eyJ6b29tIjo5Ljg3ODg2NzcxMzczNjExMiwiY2VudGVyIjpbNjMuMDQyMjQ4MjU4OTAwNDgsMzcuNDE4ODA3Mjg4OTQxNzFdLCJsYW5kc2NhcGVWaWV3Ijp0cnVlLCJyYXN0ZXJzIjp7fSwiYWN0aXZlTGF5ZXJzIjpbeyJpZCI6IjE2YjQ3MTYxNjdkLWxheWVyLTAifSx7ImlkIjoiMTZiMjM2MjlhNDItbGF5ZXItMCJ9XX0%3D

fannycc:
Sorry guys, I can't access the demo, it asks me for user and pass

weberjavi:
You have to have good pulse to toggle that one @aabdaab :D
image

aabdaab:
Actually, I found another issue, if you go to this link I posted above you'll see that without clicking around both Geodescriber and Human Pressures treemap are not displaying

fannycc:
@weberjavi thanks, here my comments;

- Checking the possible tiling options it looks like treemapSliceDice tries to split horizontally and vertically but I don't know if that will solve anything really. 
- About the opacity on the No pressure 'square' you are right sorry, all of them should have a solid fill to avoid issues. 
-My thinking is that if we don't set a minimum width and height for layers then the user won't be able to see them or click on them, so it doesn't matter that much that the visualization is loyal to reality if we can't understand it or interact with it, we already did this in CW sankeys and not sure if Trase is doing it with their. If we keep having problems with the treemap, I suggest that in the next iteration we change it, either leave it as an extra support for visual comprehension and toggle layers in a regular way, or get rid of it at all. We though it could work well as we saw them in other products. 
- If we go only for everything or nothing when showing the texts, we would have some big squares with nothing at all, the indicators name are quite large, so if we set a black or white reasoning, the chance of not having nothing at all will be big. 

I'm not sure if we've implemented a treemap in other project, but maybe worth to check with the team? Ed has told me sth about it. sorry if I'm not being too helpful

weberjavi:
Hey @fannycc great. I will fix the opacity on 'Non pressure' and will set the three different options for text display.
About the minimum visible squares seems that it needs to be some discussion and also would be great to have some testing (even from partners). Would be nice to have the input from @benlaken or @gretacv.
Do you think that could be part of the next iteration of the widget? Maybe we can set a small meeting to talk about it and about species widget once the science team has met MoL team next week?

fannycc:
That sounds perfect @weberjavi 
I think a review of the widgets we use in landscape is needed :) 

Thanks!

weberjavi:
Well this is how it is looking.
I did not managed to make the tooltip work 🤦‍♂ 
Probably @aabdaab or @kfilip94 with a fresher brain will get it to work 😅 
image

weberjavi:
@kfilip94 , @martindub don't forget to take a look at that one ☝️

kfilip94:
I'm not sure if it's possible to attach tooltip to the cells. Maybe we can borrow something from [this example](https://www.d3-graph-gallery.com/graph/treemap_custom.html), just use group labels as our area label. We can also think about [pie-chart with fixed labels](https://www.d3-graph-gallery.com/graph/donut_label.html). What do you think?

weberjavi:
@kfilip94 probably a question to @fannycc ?

fannycc:
Hey @kfilip94 @weberjavi before we start trying other approaches maybe we should ask if someone has made something similar with treemaps before. Fausto and I found this tree map in Resource Watch with tooltips -> https://resourcewatch.org/data/explore?zoom=3&lat=0&lng=0&basemap=terrain&labels=light&layers=%255B%257B%2522dataset%2522%253A%2522565c3cb1-8015-4029-b497-e9245608580f%2522%252C%2522opacity%2522%253A1%252C%2522layer%2522%253A%252291039b2b-6b72-4f01-92da-f385324e8aa0%2522%257D%255D&page=1&sort=most-viewed&sortDirection=-1&topics=%255B%2522population%2522%255D

But ideally, we were thinking in something similar to this -> https://app.flourish.studio/@flourish/hierarchy

fannycc:
Also, if it might be interesting, Ed has shown me this library when I started working in Climate Watch and it has a section called treemaps...Not sure if it can be styled to look like our design tho, I guess so. http://recharts.org/en-US/examples/CustomContentTreemap 

I hope you find something helpful.

weberjavi:
mmm, I'm started to get I bit confused right now. Not sure what are we trying to discuss here :/
The examples you shared above are quite different and I'm not sure how they resolve the 'issues' we have now in hand (the segments of the treemap not being perfectly square).
Probably all of the use d3 under the hood (recharts does) so the limitation of treemap tiling will be the same that we have now. 
We have to bear in mind that InVision designs are static and do not take into account the variability of the data that feeds the chart, so not all the cases will be as aesthetically pleasant, but they just display the data behind the chart. 
Right now I think the implementation is only laking the tooltip display when there's no label inside the treemap segment (because of space constrains), and hopefully this bug would be addressed in another PR 🤞 
If we are going to be reviewing the landscape widgets altogether I think this conversation could happen then (also about the concerns about having a minimum `width` / `height` per segment).
If there is any concern about the implementation I would be open to change whatever it needs to be changed to match your ideas/needs. 
But for open a discussion I think it would be better to do it through Slack or in a meeting and move this PR forward (or close it if we don't want to have this widget on landscape view)
😅

PR #41, opened 2019-06-12
Layer opacity bug

aabdaab:
Before we were setting layer's `opacity` directly on the WebScene's layer along with setting opacity in the url (in `handleChangeOpacity` method in `legend-component`) and it was causing issues on removing layers (we were changing only the visibility of the layers but the opacity was not resetted and we had two sources of truth).
Now, `handleChangeOpacity` is only changing opacity in the URL and then in the `arcgis-layer-manager-component` we're driving that value to feed it to webscene's layer.


kfilip94:
This bug still occurs when you change the layer to Rarity/Richness!
ohfse-6k3rv

PR #42, opened 2019-06-13
Extract common hooks to utils/hooks file

aabdaab:
I was reading about useful patterns in the `hooks` world and created a custom hook `useWatchUtils` and extracted it to `utils/hooks` file. We can identify more on the way :)


aabdaab:
good point, Javi, it slipped my eyes. I'm gonna move it to an `esri`-dedicated file in the `src/hooks` hooks folder; `utils` suggests that we store helper functions there and `src/hooks` implies that it is as important as `components`.

aabdaab:
done

PR #43, opened 2019-06-14
Feature/sidebar style tweaks

kfilip94:
This PR introduces a lot of small styles tweaks for the sidebar, to be more consistent with the design. 
Also, fixes radio-button bug - when you wanted to unselect radio button, it didn't work. I fixed this by simply changing an input event from `onChange` to `onClick`. Browser console complains about that, maybe there is a better way to fix this. We should think about extracting styles for radio-buttons/checkboxes and sidebar active-layers sections because they are repeated in a few different places, but I don't want to do it in this PR.

[PIVOTAL](https://www.pivotaltracker.com/story/show/166479396) | [INVISION](https://projects.invisionapp.com/d/main#/console/17484384/362411410/inspect)
Untitled

### DIFF
--- a/src/components/biodiversity-layers/biodiversity-layers-styles.module.scss
+++ b/src/components/biodiversity-layers/biodiversity-layers-styles.module.scss
@@ -11,7 +11,7 @@
 
 .titleSection {
   display: flex;
-  padding-left: $sidebar-paddings;
+  padding: 0 $sidebar-paddings;
 }
 
 .widgetTitle {
@@ -34,7 +34,7 @@
   font-size: $font-size-xs;
   margin-top: 0;
   margin-bottom: 14px;
-  padding-left: $sidebar-paddings;
+  padding: 0 $sidebar-paddings;
 }
 
 .fineScaleWrapper {
--- a/src/components/biodiversity-layers/biodiversity-layers-styles.module.scss
+++ b/src/components/biodiversity-layers/biodiversity-layers-styles.module.scss
@@ -2,12 +2,16 @@
 @import 'styles/settings';
 
 .wrapper {
-  padding-bottom: $sidebar-paddings;
-  padding-left: $sidebar-paddings;
+  @include backdropBlur();
+  padding-bottom: 30px;
+  &:first-child {
+    padding-top: $sidebar-paddings;
+  }
 }
 
 .titleSection {
   display: flex;
+  padding-left: $sidebar-paddings;
 }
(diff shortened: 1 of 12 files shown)


src/components/category-box/category-box-styles.module.scss
```diff
@@ -61,6 +62,7 @@
   line-height: 30px;
```
weberjavi: Maybe not related with this PR but there are some updated styles. The closed boxes default state has an arrow, do you think it is possible to add it on this PR?
image

src/components/biodiversity-layers/biodiversity-layers-styles.module.scss
```diff
@@ -2,36 +2,46 @@
...
+  @extend .widgetTitle;
+  padding-left: calc(2 * #{$sidebar-paddings})
 }
 
 .description {
   color: $alto;
   font-size: $font-size-xs;
   margin-top: 0;
   margin-bottom: 14px;
+  padding-left: $sidebar-paddings;
```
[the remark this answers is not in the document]
kfilip94: Done

src/components/biodiversity-layers/biodiversity-layers-styles.module.scss
```diff
@@ -2,36 +2,46 @@
 @import 'styles/settings';
 
 .wrapper {
-  padding-bottom: $sidebar-paddings;
-  padding-left: $sidebar-paddings;
+  @include backdropBlur();
```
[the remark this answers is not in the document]
kfilip94: Screenshot from 2019-06-14 17 17 17
Safari and Chrome Canary are not for peasants :(
But I solved the problem, Safari doesn't like `position: fixed`

src/components/category-box/category-box-styles.module.scss
kfilip94: I can change it in this PR, no problem :+1:

kfilip94: Actually, it's not that simple, because these two icons are different. The double-arrow icon has his own padding and different size, so trying to keep this in this aligned with a single arrow icon it's a nightmare. I need a coherent version of both arrow from @fannycc, I'll skip it in this PR.
Screenshot from 2019-06-14 18 08 02
Screenshot from 2019-06-14 18 07 35

kfilip94:
:point_up_2:  ready again

PR #44, opened 2019-06-17
Feature/new biodiversity facets layer

weberjavi:
This PR adds new Biodiversity Facets layer.
[PIVOTAL](https://www.pivotaltracker.com/story/show/166447966)

This layer has two main novelties compared with the 'old' `rarity-richness-GRID` layer: 
- The features of the grid layer have been dissolved to store only one feature per grid cell.
- It has embedded credentials allowing non logged in users to visit the app.

Several changes/fixes have been addressed in this PR:
- `landscape-view-manager` component has been split in two components letting the `manager` handle camera and url param update and creating a new `grid-layer` components that takes care of 'watching' view extent changes and painting the aggregated grid cells.
- There was a [known bug](https://github.com/Vizzuality/half-earth-v3/pull/40#issuecomment-501666246) when first entering the landscape mode. We where waiting for the `stationary` state of the view to look for features inside the viewport, now we wait until the `grid view layer` is not `updating` for the first time so ensuring that the features are there before querying.
- Whenever the aggregated grid cells change a couple of actions are dispatched, one to update grid cell data (related to species and human pressures) and another one to update grid cell geojson.
- The action that updates the geojson is intercepted by sagas middleware to handle the fetching of geo-describer data. 
- The firefly basemap has also been updated.

Once this PR is aproved and merged `rarity-richness-GRID` layer should be deleted from the arcgis online webscene.

Sorry for the big PR :/


aabdaab:
I noticed that the geo-describer stopped working (try couple of places it always says `Cannot load..`)
image

aabdaab:
also, maybe I'm missing something, but I still have to put credentials on the `dataGlobe`

kfilip94:
I see that now the landscape-view rectangle doesn't appear, is this because `rarity-richness-GRID` layer has been removed?

src/sagas.js
```diff
@@ -1,34 +1,82 @@
...
+// const SPECIES_FETCH_DATA_ERROR = speciesModule.actions.SPECIES_FETCH_DATA_ERROR
 
-const molAPI = 'https://api.mol.org/1.x/species/info?scientificname='
+const SET_GEO_DESCRIPTION_LOADING = geoDescriberActions.setGeoDescriptionLoading
+const SET_GEO_DESCRIPTION_READY = geoDescriberActions.setGeoDescriptionReady
+const SET_GEO_DESCRIPTION_ERROR = geoDescriberActions.setGeoDescriptionError
 
-function* watchGridCellDataFetch() {
-  yield takeLatest('SET_GRID_CELL_DATA', fetchSpeciesData)
+// const molAPI = 'https://api.mol.org/1.x/species/info?scientificname='
```
kfilip94: I assume that we want to keep this commented code here?

weberjavi: yeah, is there to eventually drive the Species widget (that needs some rethinking)

src/components/grid-layer/grid-layer-component.jsx
```diff
@@ -0,0 +1,104 @@
...
+                    // Add it to the store
+                    setGridCellGeometry(gridCellGeometry)
+                  })
+              }
+              gridCellRef.current = gridCells;
+            })
+          }
+        })
+        return function cleanUp() {
+          watchHandle && watchHandle.remove();
```
kfilip94: Does it make sense to move it to a new function? It appears two times
```
const cleanHandles = () => {
  watchHandle && watchHandle.remove();
  queryHandle && queryHandle.cancel();
}
```

src/components/grid-layer/grid-layer-component.jsx
```diff
@@ -0,0 +1,104 @@
...
+  const [gridViewLayer, setGridViewLayer] = useState(null);
+  const [gridCellGraphic, setGridCellGraphic] = useState(null);
+  // References for cleaning up graphics
+  const gridCellRef = useRef();
+
+  //Create the graphics layer on mount
+  useEffect(() => {
+    loadModules(
+      [
+        "esri/Graphic",
```
kfilip94: I was thinking, maybe we can do something with these strings here, when we do `loadModules()`, e.g. move them to separate file as constants, and then import them in our components? It's more like an open question, ofc it's not an issue for this PR
```
export const GRAPHICS_LAYER = "esri/layers/GraphicsLayer";
```

src/components/human-impact-layers/human-impact-layers-component.jsx
```diff
@@ -9,9 +9,9 @@ import { HUMAN_PRESSURE_LAYER_ID } from 'constants/human-pressures';
 
 const HumanImpactLayers = ({ map, rasters, setRasters, setLayerVisibility, activeLayers }) => {
   const humanImpactLayerActive = activeLayers.find(l => l.id === HUMAN_PRESSURE_LAYER_ID);
-  const alreadyChecked = humanImpactLayerActive && humanPressuresLandUse.reduce((acc, option) => ({ 
+  const alreadyChecked = humanImpactLayerActive && (humanPressuresLandUse.reduce((acc, option) => ({
     ...acc, [option.value]: rasters[option.value]
-  }), {}) || {};
+  }), {}) || {});
```
kfilip94: It's more a question than request, but does this last `|| {});` do anything? In case when `humanImpactLayerActive` will be false, we will return nothing instead of empty object, is that right?

weberjavi: Maybe it would be cleaner with a ternary, but it is working the same way.
Just added the parenthesis for disambiguation

src/components/grid-layer/grid-layer-component.jsx
weberjavi: done

src/components/human-impact-layers/human-impact-layers-component.jsx
kfilip94: :ok_hand:

weberjavi:
hey @kfilip94 please check this updates, I think the grid should be working now (being painted all the times).

PR #45, opened 2019-06-19
Feature/default opacity and finescaledata max scale

weberjavi:
This PR fixes the *maxScale* attribute for biodiversity fine scale layers (Hummingbirds and South Africa Layers). This feature was requested from MoL team on our last core meeting (only global biodiversity data must disappear when in landscape view).
It also sets default opacity for newly added layers to `0.6`, that has been a request from Esri agreed with the design team.
[PIVOTAL](https://www.pivotaltracker.com/story/show/166767898)

Kapture 2019-06-19 at 9 59 50


kfilip94:
:point_up: checking

PR #46, opened 2019-06-19
Feature/contentful partners update

weberjavi:
This PR adapts partners component to display the new `schema` received from the contentful API.
A new section has been added (`Platform partner`) and some logos have been updated.

Also fixed the `reversed` Human pressures legend gradient.

As an extra I've added a `typography-extends.scss` file to mimic the typography styles defined on the [InVision UI Kit](https://projects.invisionapp.com/d/main#/console/17484384/365514646/preview)
Also added some colour variables (again with the names of the UI Kit) and removed some unused font sizes variables.


src/styles/typography-extends.scss
```diff
@@ -0,0 +1,52 @@
+@import 'styles/settings';
+
+// Big titles, HE-Meter page
+%displa1 {
```
kfilip94: Isn't it a typo?

weberjavi: Well it is, is lacking a `y`
image
Will fix it.

src/components/about/partners/partners-styles.module.scss
```diff
@@ -18,25 +19,23 @@
...
-  letter-spacing: 1px;
-  line-height: 30px;
   margin: 0;
   margin-bottom: 50px;
   text-shadow: 0 0 10px 0 rgba(255,255,255,0.5);
 }
 
 .logosWrapper {
   display: flex;
+  justify-content: center;
```
kfilip94: After this change, the logos are justified to the center, but according to the design they should be justified to the left:
Screenshot from 2019-06-21 09 55 48

I'll suggest to remove it and use a theme class in `partners.js`:
image
You can pass this way a custom class and e.g. justify content to the center:
Screenshot from 2019-06-21 09 58 39

kfilip94:
@weberjavi it looks good to me :+1:  check my comments above

kfilip94:
@aabdaab I added the review changes, ready to review again

src/styles/settings.scss
```diff
@@ -24,30 +24,36 @@ $marzipan: #F7DEA1;
 $color-facebook: #39579a;
 $color-twitter: #69ace0;
 
+// Project colours
+$space-background: #09141A;
+$brand-color-main: #18BAB4;
+$brand-color-secondary: #0F2B3B;
+$dark-text: #0A212E;
+$grey-text: #D8D8D8;
+$light-text: #FFFFFF;
```
aabdaab: I noticed that couple of colours here are duplicated ($white and some others) so maybe instead of:
`$light-text: #FFFFFF;` we could initialize it like so: `$light-text: $white;`

aabdaab:
I added one comment, the rest looks fine to me :)

PR #49, opened 2019-06-21
Fix/dynamic redux modules

weberjavi:
This PR adds a `sagaRegistry` (similar to `recuderRegistry`) to allow adding sagas to the `sagaMiddleware` once they are needed (instead of all at once on the first load). This avoids the registration of the redux modules imported on those sagas.

I've added the sagas for each module under the `redux-modules` folder.
[PIVOTAL](https://www.pivotaltracker.com/story/show/166605669)


src/store.js
```diff
@@ -25,7 +26,22 @@ export default () => {
...
   );
-  sagaMiddleware.run(rootSaga)
+  // We set an event listener for the saga registry
+  // So that whenever a new saga gets added
+  // We replace the sagas with the new ones
+  sagaRegistry.setChangeListener((sagas) => {
+    function* allSagas(getState) {
+        yield all(sagas);
+    }
+    sagaMiddleware.run(allSagas);
```
sorodrigo: you should probably call cancel on the task that `sagaMiddleware.run` returns before calling run again:
```js
let task;
.
.
.
if (task) {
  task.cancel();
}
task = sagaMiddleware.run(allSagas);
```

aabdaab:
Maybe in this PR we could defer registering `metadata` reducers as well:
image

src/store.js
weberjavi: Hey @sorodrigo thank you so much for the review  🙇

weberjavi:
> Maybe in this PR we could defer registering `metadata` reducers as well
@aabdaab what do you think about deferring `metadata` and `pageTexts` modules in another PR?
 It is going to require a bigger refactor (info modal is registered on several places where modal opening actions happen, as we are importing actions from the entry file of the modal hence registering it at the same time. We should maybe rethink this pattern to avoid this registration). 
Also some refactor would be needed on the  `About` component because the button that triggers the component is part of the component itself (don't think that's the best approach) so maybe too many things to do here.

weberjavi:
:point_up: ready :)

simaob:
Don't let PRs rot!

weberjavi:
Sorry did not noticed the conflicts after the last `develop` updates :(
Fixed now. Ready again @kfilip94 , @aabdaab ☝️

src/pages/data-globe/data-globe-component.jsx
```diff
@@ -43,7 +48,7 @@ const DataGlobeComponent = ({
   const isBiodiversityActive = activeCategory === 'Biodiversity';
   const isHumanPressuresActive = activeCategory === 'Human pressures';
   const isProtectedAreasActive = activeCategory === 'Existing protection';
-
+console.log(hasMetadata, activeLayers)
```
kfilip94: Please remove this console.log

kfilip94:
It looks good :+1:  @weberjavi just remove this console.log and feel free to merge it

PR #50, opened 2019-06-26
Styles review

aabdaab:
Typography fixes in the sidebar, landscape sidebar, icons size etc.

This PR also addresses this fix: https://www.pivotaltracker.com/story/show/166926291 
Now biodiversity layers are shown on each zoom level.


src/components/category-box/category-box-styles.module.scss
```diff
@@ -1,10 +1,11 @@
 @import 'styles/ui.module';
 @import 'styles/settings';
+@import 'styles/typography-extends';
 
 .box {
   @include backdropBlur();
   box-shadow: 0 2px 10px 0 rgba(23, 27, 48, 0.15);
-  padding: 0 30px 0 20px;
+  padding: 0 20px 0 20px;
```
kfilip94: Change to `padding: 0 20px`

kfilip94:
Hover doesn't work on zoom & location widget:
Screenshot from 2019-06-26 13 35 12
The reason why is because you have restored SVG icons to the previous version. One important difference between original and  modified files was removing every `fill` or `stroke` properties from child nodes - that allows you to color SVG from CSS level on hover

fannycc:
These buttons don't have a hover state if you check the UI KIT. We only have active states for location and full screen. The colored version of the icons is included. On hover we should show the tooltips showing the meaning of them: 
Captura de pantalla 2019-06-26 a las 14 44 49

fannycc:
icons also available here -> https://www.dropbox.com/sh/6lazr7dv9d3eckq/AADTNS2IvFVgZpWIuKr-H7Sba?dl=0

kfilip94:
@fannycc is right :+1:  I was wrong :-1:  sorry :disappointed:

PR #54, opened 2019-06-27
Human pressures title

aabdaab:
`Human pressures` title was cropped in Firefox so I increased the fixed height for that one as well. Maybe in the future we coud investigate how to best dynamically adjust fixed height depending on the text length but that might be difficult.


kfilip94:
Firefox:
Screenshot from 2019-06-27 10 22 19
but
Google Chrome:
Screenshot from 2019-06-27 10 21 13
There is a gap between the title and bottom line on Chrome

PR #59, opened 2019-06-28
Feature/google analytics

kfilip94:
This PR implements Google Analytics actions for these scenarios: 
- user adds a layer,
- user removes a layer,
- user removes a layer via legend,
- user opens layer info modal,
- user opens half earth meter modal,
- user opens about page,
- user changes the layer opacity,

[PIVOTAL](https://www.pivotaltracker.com/story/show/166605143)

### DIFF
--- a/src/components/checkbox-group/checkbox/checkbox.js
+++ b/src/components/checkbox-group/checkbox/checkbox.js
@@ -1,17 +1,21 @@
 import React from 'react';
 import { connect } from 'react-redux';
 import Component from './checkbox-component';
-import actions from 'redux_modules/metadata';
+import metadataActions from 'redux_modules/metadata';
+import * as googleAnalyticsActions from 'actions/google-analytics-actions';
+
+const actions = { ...metadataActions, ...googleAnalyticsActions };
 
 const CheckboxContainer = props => {
 
   const handleInfoClick = (option) => {
-    const { setModalMetadata } = props;
+    const { setModalMetadata, openLayerInfoModalAnalyticsEvent } = props;
     setModalMetadata({
       slug: `${option.slug}`,
       title: `${option.name} metadata`,
       isOpen: true
     });
+    openLayerInfoModalAnalyticsEvent({ slug: `${option.slug}` });
   };
  return (
   <Component
--- a/src/components/human-impact-layers/human-impact-layers-component.jsx
+++ b/src/components/human-impact-layers/human-impact-layers-component.jsx
@@ -7,7 +7,7 @@ import MultipleActiveLayers from 'components/multiple-active-layers';
 import { humanPressuresLandUse } from 'constants/human-pressures';
 import { HUMAN_PRESSURE_LAYER_ID } from 'constants/human-pressures';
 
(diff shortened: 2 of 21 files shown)


aabdaab:
this bit: `user opens layer info modal` works only via Legend, it should also work via Sidebar

aabdaab:
when user opens Half-Earth modal there's an error in the console related to google analytics:
image

kfilip94:
:point_up: Ready again

weberjavi:
Hey @kfilip94 does it makes sense to add `WIP` label while you're still working on the missing analytics?

kfilip94:
I thought we want to merge this part for now, but I can finish other tasks in this PR as well, @simaob?

kfilip94:
image

src/components/checkbox-group/checkbox/checkbox.js
```diff
@@ -1,17 +1,21 @@
 import React from 'react';
 import { connect } from 'react-redux';
 import Component from './checkbox-component';
-import actions from 'redux_modules/metadata';
+import metadataActions from 'redux_modules/metadata';
+import * as googleAnalyticsActions from 'actions/google-analytics-actions';
```
weberjavi: If we are only using `openLayerInfoModalAnalyticsEvent` wouldn't it be better to:
`import { openLayerInfoModalAnalyticsEvent } from 'actions/google-analytics-actions';`
 and
 `const actions = { ...metadataActions, openLayerInfoModalAnalyticsEvent };`?

src/components/human-impact-layers/human-impact-layers-component.jsx
```diff
@@ -27,6 +28,11 @@ const HumanImpactLayers = ({ map, rasters, setRasters, setLayerVisibility, activ
 
     const mosaicWhereClause = `Name IN('${rasterNames.join("','")}')`;
 
+    const analyticsParmas = { slug: option.slug, query: { viewMode: VIEW_MODE.LANDSCAPE }};
```
weberjavi: Should it be `analyticsParams` instead of `analyticsParmas`?

src/components/radio-group/radio-group.js
```diff
@@ -2,17 +2,21 @@ import React from 'react';
 import { connect } from 'react-redux';
 import { capitalize } from 'lodash';
 import Component from './radio-group-component';
-import actions from 'redux_modules/metadata';
+import metadataActions from 'redux_modules/metadata';
+import * as googleAnalyticsActions from 'actions/google-analytics-actions';
```
weberjavi: The same for this, maybe importing the methods we are using instead of all the actions.
It might have some performance and bundle benefits:
https://stackoverflow.com/questions/45305272/is-there-a-performance-difference-between-import-as-from-lodash-and-import

weberjavi:
👏

weberjavi:
My only doubt is about placing the analytics middleware inside the `providers` folder. But this is a conversation that we can have later.
@kfilip94 , @aabdaab ^^

PR #61, opened 2019-07-02
Fix/mix of frontend fixes

kfilip94:
This PR fixes a few smaller front-end issues:
[PIVOTAL](https://www.pivotaltracker.com/story/show/167022520)
1. Changes background and opacity of the legend and the sidebar to 0.7:
**OLD BACKGROUND:**
old
**NEW BACKGROUND**
new

2. Fix `overflow: hidden` bug on Safari:
image (1) (1)

 and behavior of half-earth-meter globe, when we are navigating through the North/South Pole:
aoodk-dowsc

~~3. Adds missing background color after hover on FIND PLACES button:~~
Screenshot from 2019-07-02 12 04 38


fannycc:
All looks good to me. I just have one comment about the hover state in find places. Can I see it working? I'm not sure if it looks good when we interact with it.. I know it's in the UI KIT but I'd like to see it, a video would be enough :) The other option would be to leave it as it was

kfilip94:
@fannycc [here is the link](https://drive.google.com/open?id=1jAzb9Gb3P8cv-CRu1fPoDqyvhLQyMUnx)

fannycc:
@kfilip94 Thanks a lot for the video! Can we leave it as it was before? no hover, just the hand cursor. I review it with @faustoperez and it doesn't look that well working. Sorry, sometimes we need to test some things.

kfilip94:
@fannycc Done :man_factory_worker:

weberjavi:
Starting the review ☝️ 
But @kfilip94 shouldn't this PR be against `develop` and not `master`?

kfilip94:
@weberjavi it should :man_facepalming:  thanks

PR #63, opened 2019-07-02
layer ordering

weberjavi:
This PR adds layers reordering functionality through the legend.
[PIVOTAL](https://www.pivotaltracker.com/story/show/166604528)

The reordering logic has been split in two; The legend handles the update of the URL params with the new order and layer reordering on the map is handled by the `layer-manager` (based on URL `activeLayers` order).

This is the first iteration and some bits would be tweaked in later PRs:
- Human pressures ordering is not working at all. We will need to investigate if this is due to the nature of the layer (`imageryLayer`). There's already an [open bug](https://www.pivotaltracker.com/story/show/166926340) related to this.
- Also some PR to vizzuality components repo should be made in order to [allow dragging the whole legend item](https://www.pivotaltracker.com/story/show/166925130) to reorder it (instead of being able to reorder them by dragging on the dedicated icon).

About this last point (removing the drag icon I wonder if it is going to be clear to the user that the legend is draggable without it or without any hint to let them know this functionality). Maybe something to add to the eventual 'tour/tutorial'? @fannycc , @faustoperez ?

Kapture 2019-07-03 at 16 23 01


weberjavi:
@kfilip94 , @aabdaab ☝️ ready to review.
@fannycc  👩‍🎨  @faustoperez 👨‍🎨 there's a question for you on the PR description

fannycc:
@weberjavi yup, adding this feature to the tutorial guide sounds good to me. Actually the reason why the user needed to grab an icon to drag instead of the big legend was a question I asked during one design meeting a while ago and there weren't any reasons. For me the easiest thing from an user perspective would be being able to grab the whole box. The current icon is not that intuitive for all the audience and no explanations are made there about it either. I feel is more comfortable if the user don't need to put the cursor in a particular small icon to drag the legend. Any thought? @faustoperez

faustoperez:
If we remove the drag icon we can always change the cursor over the draggable area

`cursor: move;`

:)

fannycc:
yup! agree with @faustoperez about the cursor :)
I've talked with @kfilip94 in Porto about the hand cursor we could use here. Let us know if you need us to review it :)

weberjavi:
Great, thanks @fannycc & @faustoperez. I think is a nice solution.
Once the vizzuality components PR is merged will let you know :)

kfilip94:
@weberjavi Looks and works great, thank you for implementing this :1st_place_medal: 
My only concern is that maybe we should disable drag for the human pressures layer for now? If we can't drag it, it should be always on the top. Does it make sense for now?

weberjavi:
@kfilip94 I'm afraid this is a change that should be made at Vizzuality Components (as for now all the legend items are draggable whenever the `sortable` attribute is true). We might try to hide the drag icon on the human pressures item but that would not avoid it to change order when other items are dragged over it.

kfilip94:
I'd leave it as it is then, hiding the drag icon would be also confusing. Thanks :+1:

aabdaab:
I'd like to draw your attention to this bug; when you activate a layer and then click on an info button it no longer works and app fails:
image

aabdaab:
I think this is a good moment to start talking about introducing frontend tests to half-earth! let's discuss this tomorrow 🙌

PR #65, opened 2019-07-04
Bug/filter marine grid cells pressures

weberjavi:
This PR avoids Marine Grid cells to be accountable while calculating terrestrial human pressures (that was causing the skew of the pressures percentages).

[PIVOTAL](https://www.pivotaltracker.com/story/show/167037050)

@fannycc, @faustoperez , @gretacv, @benlaken  would it make sense to change the wording on the human pressures widget to aware users that the pressures displayed there are only terrestrial? I think eventually we would need some place to display marine pressures.


weberjavi:
Fair enough 👍

PR #66, opened 2019-07-05
Revert "layer ordering"

weberjavi:
Reverts Vizzuality/half-earth-v3#63

Gonna check on `layers-reordering` branch what's causing the info buttons bug:
https://github.com/Vizzuality/half-earth-v3/pull/63#issuecomment-508544090

### DIFF
--- a/src/components/arcgis-layer-manager/arcgis-layer-manager-component.jsx
+++ b/src/components/arcgis-layer-manager/arcgis-layer-manager-component.jsx
@@ -1,33 +1,13 @@
-import { includes } from 'lodash';
-import { LEGEND_FREE_LAYERS } from 'constants/layers-groups';
-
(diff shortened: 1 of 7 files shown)


weberjavi:
@aabdaab , @kfilip94 not sure why this error is being triggered but it is working fine locally. The thing is that on metadata reducers initial state the `data` field is set to an empty object but on the `demo`site (that has `layer-reordering` branch merged) `metadata` initial state `data` is set to `null`.
On my local `develop` branch (that do not have the `layer-reordering`branch merged) alsois working fine 🤔 
Will try to sort this out.

weberjavi:
On the browser the initial state `data`is set as `null`
image

weberjavi:
Ok, found the ofender PR (it was the dynamic redux modules on). 
Closing that one and opening a PR with the fix :)

PR #70, opened 2019-07-05
Add autofocus to search input

kfilip94:
This tiny PR adds autofocus to the location search input.
[PIVOTAL](https://www.pivotaltracker.com/story/show/167036547)

Please leave a comment if you have any concern about my solution below

### DIFF
--- a/src/components/widgets/search-widget/search-widget.js
+++ b/src/components/widgets/search-widget/search-widget.js
@@ -42,9 +42,13 @@ const SearchWidget = ({ view, openPlacesSearchAnalyticsEvent, searchLocationAnal
     handleCloseSearch();
   }
 
+  async function addSearchWidgetToView() {
+    await view.ui.add(searchWidget, "top-left");
+  }
+  
   useEffect(() => {
     if( searchWidget ) {
-      view.ui.add(searchWidget, "top-left");
+      addSearchWidgetToView().then(() => document.querySelector(".esri-search__input").focus());
       document.addEventListener('keydown', keyEscapeEventListener);
       searchWidget.viewModel.on("search-start", handleSearchStart);
       searchWidget.watch('activeSource', function(evt) {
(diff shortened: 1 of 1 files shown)


src/components/widgets/search-widget/search-widget.js
```diff
@@ -33,9 +32,13 @@ const SearchWidget = ({ view }) => {
     setSearchWidget(null);
   }
   
+  async function addSearchWidgetToView() {
```
aabdaab: use `es6` arrow function syntax: 
```
const addSearchWidgetToView = async () => {
 await view.ui.add(searchWidget, "top-left");
}
```

src/components/widgets/search-widget/search-widget.js
```diff
@@ -33,9 +32,13 @@ const SearchWidget = ({ view }) => {
...
+    await view.ui.add(searchWidget, "top-left");
+  }
+  
   useEffect(() => {
     if( searchWidget ) {
-      view.ui.add(searchWidget, "top-left");
+      addSearchWidgetToView().then(() => document.querySelector(".esri-search__input").focus());
```
aabdaab: alternatively, not to use `then` syntax you could do:
```
const addSearchWidgetToView = async () => {
    await view.ui.add(searchWidget, "top-left");
    document.querySelector(".esri-search__input").focus();
};
addSearchWidgetToView();
```
inside a hook. You cannot wrap the whole useEffect with async function (`useEffect(async () => { ... })`) because `useEffect` has to return a clean up function but what you can do is define a named inner function (addSearchWidgetToView) that you will then call inside useEffect; it will be run asynchronously but the statements inside `addSearchWidgetToView` itself will be executed synchronously (so you don't have to worry that you will call `focus` on not yet existing node.

kfilip94:
:point_up: @aabdaab thanks for comments, ready again

PR #72, opened 2019-07-10
Feature/terrain exaggeration

weberjavi:
This PR adds a terrain elevation layer. That has been a design request since the beginning of the project (it was requested again while going through landscape features labels task review).

It increases the draping sensation of grid cells on the terrain.

image

It also eases the understanding of some values distributions related with altitude on some datasets
image

The factor of exaggeration could be modified through a prop on the component.
```js
<TerrainExaggerationLayer exaggeration={3}/>
```


weberjavi:
@aabdaab , @kfilip94 do not forget that this is around ☝️ :)

PR #75, opened 2019-07-11
Set up testing environment - jest configs

aabdaab:
This PR adds `jest configuration` and prepares testing environment for HE.
Couple of takeaways from setting up `jest`:
- we didn't have to add `jest` as a dependency as it's a given thanks to `react-app-rewired`
- add `__tests__` folder to src - it's a default one, jest will look for test files in here.
- add babel plugins and `.babelrc.js` config file - webpack is not doing its work in `__tests__` folder. For that reason we need babel plugins to have these js files be transformed by babel. That's also why we've got `moduleNameMapper` in jest config in package.json
- `jest` doesn't recognize anything but `.js`, so we need to mock our svgs, jpgs and CSS. Code responsible for that:
```
"\\.svg": "<rootDir>/test/mocks/svg-mock.js",
"\\.module\\.scss$": "identity-obj-proxy",
"\\.scss$": "<rootDir>/test/mocks/style-mock.js"
```
for mocking css modules we're using this useful dependency `identity-obj-proxy`. 
- two scripts to run tests in `package.json`:
```
"test": "jest",
"test:watch": "jest --watch",
```
run `yarn test` to fire the tests. `yarn test:watch` is a useful command as it watches your source code for changes and fires only the tests that are impacted by your changes.
- added `@testing-library/jest-dom` and `@testing-library/react` as useful wrappers on jest assertion methods and react-testing-library for easy DOM querying (check: `getByText`, `getByTestId`, docs: https://testing-library.com/docs/react-testing-library/intro 🤓 )
- added `firstTest.js` file just as an example, we can delete it later, it's just for you to see how you can use `jest` and `react-testing-library`
- added one snapshot; exclaimer about that, snapshots **shouldn't be used as a sole testing method**. could be useful when we add some `snapshotsSerializers` to our jest config in package.json for mapping out css styles

Any questions regarding this, please comment below, I can give further explanation! 

Also, a link to my notes: https://www.notion.so/vizzuality/Tests-3c04c83f79d74c398e2eaf143136c818


weberjavi:
The config is triggering an error when I try to make a change to `category-box-component`.
image

To reproduce the error:
- run `yarn jest --watch`
- Go to `category-box-component.jsx` file and comment line `18` (the line with the `p` element with `catgory` text inside).

@aabdaab any clue why this could be happening?

weberjavi:
`yarn jest` is working fine thought.

aabdaab:
Thanks for catching that; it's happening because `directory-named-webpack-plugin` plugin doesn't work in the test environment, we need to add our aliases ourselves inside `moduleNameMapper` jest's property in `package.json`.  `moduleNameMapper` works like a regex so it matches the first matching encounter in that object - so the order MATTERS here. You had an error because it matched that regex `"components/(.*)/(.*)$": "<rootDir>/src/components/$1/$2.jsx",` and mapped to `components/widgets/location-widget.js` (doesn't exist) while it's supposed to be `components/widgets/location-widget/location-widget.js`. The `$1` property evaluates to whateves comes out from the first `(.*)` encounter in the regex, and same goes for `$2`, `$3` etc (if you've got respectively 2, 3 `(.*)` in your regex).
It's fixed now, however I don't know why this error was thrown cause in the existing tests we don't reference location widget anywhere, nor the components that use that component within.....

PR #77, opened 2019-07-16
Set up cypress

aabdaab:
This PR adds `cypress` dependency to our codebase that will allow us adding e2e tests 🎉 
- adds `cypress` dependency
- adds `@testing-library/cypress` dependency with some utility functions (similar to react-testing-library utils, i.e. `queryByText ` command. Docs: https://testing-library.com/docs/cypress-testing-library/intro)
- adds `npm-run-all` dependency to be able to run multiple scripts in parallel (handy if you want to test as you develop)
- adds couple of scripts to `package.json`: 
  `test:e2e:dev` - to use locally. Runs both the app and cypress at the same time
  `test:e2e:ci` - for CI, we'll have to add it to Travis config
  `pre:test:e2e:ci` - to use locally if we want to imitate CI locally

The purpose of adding cypress is to have a tool that would test the whole experience a user would have; hitting the API included. It raises a question why do we need both: integration && e2e testing if we can only use cypress. I don't have a strong opinion about it, I'd like to use both to see which ones sit better. According to Kent C. Dodds, the developer who introduced `react-testing-library` the distinction between integration and e2e is in the nature of the tests:
- every time I want to make a server request - `e2e`
- if I don't need to hit the server to test a functionality - `integration`

A small diagram to help with grasping the idea:
AC00A2F5-B896-497E-9C49-F606CF93505F


weberjavi:
checking ☝️

aabdaab:
Hey Javi, thanks for the suggestions! It has a better structure now. I also added first two tests in cypress. Run `yarn cy:open` to see them in action.

PR #79, opened 2019-07-17
Feature/add info tooltips

kfilip94:
This PR adds tooltips to share, info layer, fullscreen and legend toolbox buttons. This is a temporary solution for a legend until we will implement a more customizable version of it.
This PR also removes the location marker from the map.
 [PIVOTAL - location marker](https://www.pivotaltracker.com/story/show/166924833) | [PIVOTAL - tooltips](https://www.pivotaltracker.com/story/show/166977629)

**NOTE:** Each button has a `cursor: pointer` property on hover, but on screens below you can see an arrow - this is an issue of screenshot software, so don't worry, the pointing-hand cursor is there :)

**Fullscreen button:**
Screenshot from 2019-07-17 18 14 47

**Share button:**
Screenshot from 2019-07-17 18 15 36

**Layer info button:**
Screenshot from 2019-07-17 18 15 47

**Legend buttons:**
image
image
image


kfilip94:
@fannycc I had to change it again because the position of the opacity value was also justified to the right:
image

But then I realized that I can do another workaround so now it works, good catch!
zcsnp-50ok6

fannycc:
@weberjavi it's on purpose. The location icon is very intuitive, my logic tells me that if we put a tooltip for it we should add a tooltip to the zoom in and zoom out and that's weird... It's true that might look a bit inconsistent this way. I'm not sure if I prefer to test it as it is now and make proper changes later. What do you think?

kfilip94:
I'll merge it for now as it is, we can change it in a different PR if we need :)

PR #80, opened 2019-07-22
Post robot

aabdaab:
Move to Post Robot listeners to the main file. I will leave the map-iframe files for now, until the stories are aligned with the change.

Fyi. everything is interactive now (on map-iframe it isn't, legend not responsive, sidebar hidden etc.). It works on `/v2` and `/dataGlobe` endpoints (both are pointing to the same root files)


aabdaab:
ready again ☝️

PR #83, opened 2019-07-24
Loading spinners

aabdaab:
PT task: https://www.pivotaltracker.com/story/show/167316474

Implement loading spinners to improve UX.
Identified scenarios:
- first landing on the page
- laying out layers on the globe

Check the gif below. I'm not sure if the 3D spinner looks ok next to the 3D globe, might be a bit too much. Let me know your thoughts.
screencast 2019-07-29 14-35-06

### DIFF
--- a/src/components/globe/globe-component.jsx
+++ b/src/components/globe/globe-component.jsx
@@ -19,7 +19,6 @@ const GlobeComponent = ({ sceneId, sceneSettings, onLoad, children, loadElement 
 
   return (
     <WebScene
-      style={{ backgroundColor: '#0A212E' }}
       className={cx(styles.sceneContainer)}
       id={sceneId}
       onLoad={handleOnLoad}
--- a/src/components/globe/globe-component.jsx
+++ b/src/components/globe/globe-component.jsx
@@ -11,9 +11,10 @@ const GlobeComponent = ({ sceneId, sceneSettings, onLoad, children, loadElement 
 
   const handleOnLoad = (map, view) => {
     setMap(map);
+    map.ground.surfaceColor = '#0A212E';
     setView(view);
     setLoaded(true);
-    onLoad && onLoad(map, view);
+    onLoad && onLoad(map, view);  
   }
 
   return (
--- a/src/components/globe/globe-component.jsx
+++ b/src/components/globe/globe-component.jsx
@@ -18,6 +18,7 @@ const GlobeComponent = ({ sceneId, sceneSettings, onLoad, children, loadElement 
 
   return (
     <WebScene
+      style={{ backgroundColor: '#0A212E' }}
       className={cx(styles.sceneContainer)}
       id={sceneId}
       onLoad={handleOnLoad}
--- a/src/utils/layer-manager-utils.js
+++ b/src/utils/layer-manager-utils.js
@@ -46,7 +46,7 @@ export const layerManagerOrder = (datasets, activeLayers, callback) => {
   callback({ activeLayers: updatedLayers });
 };
 
-export const createLayer = (layer) => {
+export const createLayer = (layer, map) => {
   return loadModules(["esri/layers/WebTileLayer"]).then(([WebTileLayer]) => {
     const { url, slug } = layer;
     const tileLayer = new WebTileLayer({
@@ -55,14 +55,8 @@ export const createLayer = (layer) => {
       id: slug,
       opacity: DEFAULT_OPACITY
     })
-    return tileLayer;
+    map.add(tileLayer);
   });
-}
-
-export const addLayer = (layer, map, view, handleGlobeUpdating) => {
-  handleGlobeUpdating(true);
-  map.add(layer)
-  handleLayerRendered(view, layer, handleGlobeUpdating);
 }
 
 export const handleLayerRendered = (view, layer, handleGlobeUpdating) => {
--- a/src/utils/layer-manager-utils.js
+++ b/src/utils/layer-manager-utils.js
@@ -46,7 +46,7 @@ export const layerManagerOrder = (datasets, activeLayers, callback) => {
   callback({ activeLayers: updatedLayers });
 };
 
-export const createLayer = (layer, map) => {
+export const createLayer = (layer) => {
   return loadModules(["esri/layers/WebTileLayer"]).then(([WebTileLayer]) => {
     const { url, slug } = layer;
     const tileLayer = new WebTileLayer({
@@ -55,6 +55,20 @@ export const createLayer = (layer, map) => {
       id: slug,
       opacity: DEFAULT_OPACITY
     })
-    map.add(tileLayer);
+    return tileLayer;
   });
+}
+
+export const addLayer = (layer, map, view, handleGlobeUpdating) => {
+  handleGlobeUpdating(true);
+  map.add(layer)
+  handleLayerRendered(view, layer, handleGlobeUpdating);
+}
+
+export const handleLayerRendered = (view, layer, handleGlobeUpdating) => {
+  view.whenLayerView(layer).then(mapLayerView => {
+    mapLayerView.watch("updating", updating => {
+      if(!updating) { handleGlobeUpdating(false) }
+    })
+  })
 }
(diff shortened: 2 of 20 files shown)


fannycc:
I'm either not sure if it's works good or not but also bc I'm not able to see it properly, those layers load quickly. I wouldn't expect to see there a spinner with that speed. Do we have some settings to tell the spinner when to appear? I guess it should show up only in the cases when the data takes more than 2 seconds to appear? 

About the spinner itself we can explore other options of course :)

fannycc:
oh, I'm seeing better the video and the human pressures take time yes. I'm missing the text of "fetching data" or something similar

aabdaab:
The loading time depends on the machine on which you're running the platform and the network speed as well. It also depends on whether you already rendered these layers within the current session or not; sometimes it takes 2sec or more (8sec at most so far if I remember correctly), sometimes a bit less ( 🎉) but we can't really rely on it. It's safe to say that layers take the longest to render so even for someone with good internet it's better to show him the spinners, even if it takes a second. Right now, the settings are:
- If you're activating a layer (doesn't matter which), show a spinner (note that on removing a layer we're not showing a spinner as it happens instantly)
- Watch the property from arcgis to see if this layer is updating -> whenever it stops updating (it means for us that it has rendered), remove the spinner.

aabdaab:
I can deploy this branch to half-earth-demo so you can click around!

fannycc:
@aabdaab I don't know what others think but it sounds good to me deploying as it is now to see the performance :)

aabdaab:
ok, here you go: https://half-earth-demo.vizzuality.com/dataGlobe

src/components/globe/globe-component.jsx
```diff
@@ -11,13 +11,15 @@ const GlobeComponent = ({ sceneId, sceneSettings, onLoad, children, loadElement
...
+    map.ground.surfaceColor = '#0A212E';
     setView(view);
     setLoaded(true);
-    onLoad && onLoad(map, view);
+    onLoad && onLoad(map, view);  
   }
 
   return (
     <WebScene
+      style={{ backgroundColor: '#0A212E' }}
```
kfilip94: I'm not sure, could you explain what this `backgroundColor` does in this case? Thanks :)

weberjavi:
Hey @aabdaab another scenario where this could be useful is after landing on landscape mode while we are computing the aggregated grid cell, it takes sometime for the cell to be painted on the map, what do you think?

src/utils/layer-manager-utils.js
```diff
@@ -46,7 +46,7 @@ export const layerManagerOrder = (datasets, activeLayers, callback) => {
   callback({ activeLayers: updatedLayers });
 };
 
-export const createLayer = (layer, map) => {
+export const createLayer = (layer) => {
```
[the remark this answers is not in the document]
fannycc: I'd prefer the spinner not to show up when removing a layer. I liked Marta's explanation and it makes sense to me--> "...removing a layer we're not showing a spinner as it happens instantly)" otherwise it can be overwhelming.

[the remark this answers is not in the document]
weberjavi: Yeah I agree, indeed my argument was that we might not need it at all when adding `Biodiversity` layers, because it is quite fast and the user already is seeing that something is happening because the layer is being painted (something that does not happen when adding the `Human pressures` rasters, as some computation needs to be done after `clicking` the button to add the layer).
Does this make sense?
My point is: I would remove the spinner on `Biodiversity` layers creation

src/components/globe/globe-component.jsx
aabdaab: I set the background color to prevent the user from seeing a naked globe :)

src/utils/layer-manager-utils.js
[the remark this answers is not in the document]
fannycc: Agree with you @weberjavi

src/components/globe/globe-component.jsx
fannycc: Talking about seeing a globe "naked" is there any way we can't avoid showing the satellite basemap when we land? I remember having some conversation with Javi about this. Could we leave the loading screen until the custom basemap is loaded?

kfilip94: @aabdaab Well, I think it doesn't work that way, because this will set the background color of the loading screen, you're setting the background color of the globe here:
image
I 'd suggest to remove `style={{ backgroundColor: '#0A212E' }}` from the WebScene and wrap the `loadElement` element into div with dark background color.

weberjavi: So @fannycc I think the best solution for us would be to use the Firefly as `basemap` that way we would avoid fetching the tiles for the `satellite imagery` (that we are currently using as `basemap`) and the Firefly when we just want to display the Firefly. That would have two main benefits for us, one would be that we wouldn't see this rendering of the Firefly over the satellite `basemap` and the other one is that we would increase the performance of the `map` as we will only be `fetching` one tile instead of two for the needed globe area (not sure if this performance increase would be perceptible by the user).
But for that we would need to create tiles for the higher zoom levels, and not sure if that's something we can do (maybe @benlaken or @gretacv have something to say here) as has been J.Nelson the one that has been creating the tiles so far.

weberjavi: On this image you can see that we are fetching the satellite imagery although we are just _covering_ it with the vibrant map tiles
image

src/utils/layer-manager-utils.js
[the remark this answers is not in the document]
aabdaab: That's true, biodiversity layers are rendering way faster than the other ones, we can remove the spinner there. It's a bit slower when you choose from the Fine Scale Data but still, there is something going on the screen - going to the location, rendering tiles.

src/components/globe/globe-component.jsx
fannycc: whatever works best :)

aabdaab:
Thanks all for your input!
- I added a commit that removes spinners on biodiversity layers ✅ 
- About having no animation of landscape sidebar on leaving landscape mode: I checked and it's happening on production as well, so I suggest fixing it in a different PR
- `seems that some actions are triggered more than once (isGlobeUpdating: false)` - I tried fixing that with checking `isGlobeUpdating` query param `if(!updating && isGlobeUpdating) { handleGlobeUpdating(false) }` but that was not working as expected, it wouldn't remove the spinner at all. It might be happening because of its placement in the `watcher`.
- Spinner when computing aggregated grid cells --> I agree, will add that.

aabdaab:
Another thing we also need to add is a spinner when enabling human pressure layers through the widget on the landscape mode

aabdaab:
Ready for re-review!
- I added spinner when computing aggregated grid cells ✅ 
- Spinner when enabling human pressure layers from the widget on landscape mode ✅ 
- Rebased with develop, merged the changes about switching between experiences ✅

kfilip94:
I found a small issue when we are moving the viewport in landscape mode, then after quick zoom out, before the data will load, we end up with infinite loading spinner:
gn0c3-u9fcr

It persists until we trigger another loading action

kfilip94:
I found a small issue when we are moving the viewport in landscape mode, then after quick zoom out, before the data will load, we end up with infinite loading spinner:  
gn0c3-u9fcr
It persists until we trigger another loading action

aabdaab:
Thanks for spotting that, I put the handler in the cleanup function as well and can't reproduce it anymore.

kfilip94:
@aabdaab thanks :+1:

aabdaab:
@weberjavi check out the changes on Monday and if you spot something feel free to change it in this PR and merge it.
To sum up, from the last review, these are the changes:
- I added a commit that removes spinners on biodiversity layers ✅
- I added spinner when computing aggregated grid cells ✅
- Spinner when enabling human pressure layers from the widget on landscape mode ✅
- Rebased with develop, merged the changes about switching between experiences ✅
About having no animation of landscape sidebar on leaving landscape mode: I checked and it's happening on production as well, so I suggest fixing it in a different PR.

About your comment `Seems that some actions are triggered more than once (isGlobeUpdating: false)` - I tried fixing that with checking `isGlobeUpdating` query param in `handleLayerRendered` handler:

```if(!updating && isGlobeUpdating) { handleGlobeUpdating(false) }```

but that was not working as expected, it wouldn't remove the spinner at all. It might be happening because of its placement in the watcher.

PR #84, opened 2019-07-25
Feature/back button in landscape sidebar

kfilip94:
This PR adds the back button to the landscape sidebar. One issue here is the delay, nothing is happenig for 2 seconds before the animation starts, actually I don't know how to solve it for now. The only alternative I found is that I can change the zoom immediately, without animation, but this is not smooth at as well. @fannycc what do you think?
[PIVOTAL](https://www.pivotaltracker.com/story/show/166925607)

CLICK -> delay + animation:
xtm3h-or259

CLICK ->  immediate zoom change + delay + animation:
xd58m-q3xgd


kfilip94:
@fannycc sure :+1: I'll freeze this PR until then, thanks :)

fannycc:
@kfilip94 I also checked the older platform and the close landscape is very smooth there. I know it might not be helpful as we're playing with other technology but just in case :)

src/components/landscape-sidebar/geo-description-widget/geo-description-widget-component.jsx
```diff
@@ -40,4 +48,4 @@ const GeoDescriptionWidget = ({ data, loading, error, view }) => {
   );
 };
 
-export default GeoDescriptionWidget;
+export default connect(null, actions)(GeoDescriptionWidget);
```
weberjavi: Please do not `connect` the `component`. Following the `container`/`component` pattern we usually make the `component` as 'dumb' as possible, just passing the needed props from the `container` to the `component` and leaving the latest just for rendering purposes.

src/components/landscape-sidebar/geo-description-widget/geo-description-widget-component.jsx
```diff
@@ -1,10 +1,19 @@
...
+import * as actions from 'actions/url-actions';
+import { connect } from 'react-redux';
+import FixedHeader from 'components/fixed-header'
+
+const GeoDescriptionWidget = (props) => {
+  const { data, loading, error, view, changeGlobe } = props;
+
+  const handleBackClick = () => {
+    const params = { zoom: 7 }
+    view.goTo(params).then(() => changeGlobe(params));
```
weberjavi: We do not need to change the `zoom` param on the `URL` after the `goTo` method has been resolved. We are already updating the `zoom` param whenever a change on the `view.zoom` happens through a watcher on the `zoom-widget` component. So we do not really need the `changeGlobe` action here.

src/components/landscape-sidebar/geo-description-widget/geo-description-widget-component.jsx
```diff
@@ -1,10 +1,19 @@
...
 import styles from './geo-description-widget-styles.module.scss';
+import * as actions from 'actions/url-actions';
+import { connect } from 'react-redux';
+import FixedHeader from 'components/fixed-header'
+
+const GeoDescriptionWidget = (props) => {
+  const { data, loading, error, view, changeGlobe } = props;
+
+  const handleBackClick = () => {
+    const params = { zoom: 7 }
```
weberjavi: If we want to make this method perform faster we can pass also the `center` of the actual view as a param. We can get the actual `center` from the `globe` `key` of the `URL`. Not really sure why it performs better but it does (the `center` is an optional parameter but seems that explicitly passing it to the method eases things out, maybe that way it has not to compute the actual `center` or `target` for the method :)

src/components/landscape-sidebar/geo-description-widget/geo-description-widget-component.jsx
```diff
@@ -1,10 +1,19 @@
...
 import { VIEW_MODE } from  'constants/google-analytics-constants';
 import styles from './geo-description-widget-styles.module.scss';
+import * as actions from 'actions/url-actions';
+import { connect } from 'react-redux';
+import FixedHeader from 'components/fixed-header'
+
+const GeoDescriptionWidget = (props) => {
+  const { data, loading, error, view, changeGlobe } = props;
+
+  const handleBackClick = () => {
```
weberjavi: Building on the idea behind `container`/`component` pattern would be nice if you can extract the event handler (`handleBackClick`) outside of the `component` passing it as a `prop` from the `container`.

src/components/landscape-sidebar/geo-description-widget/geo-description-widget-component.jsx
```diff
@@ -30,8 +39,7 @@ const GeoDescriptionWidget = ({ data, loading, error, view }) => {
 
   return (
     <div className={styles.container}>
-      <ShareModal theme={{ shareButton: styles.shareButton }} view={view} viewMode={VIEW_MODE.LANDSCAPE} />
-      <h2 className={styles.title}>{data.title}</h2>
+      <FixedHeader closeSidebar={handleBackClick} title={data.title} view={view} viewMode={VIEW_MODE.LANDSCAPE} autoHeight={true} />
```
weberjavi: We might need to add this header to the `error` case to allow users to exit landscape mode also when an error occurs on the geo describer API. We might want to double check with @fannycc what is the desired `title`.

kfilip94:
@weberjavi Thanks a lot for your feedback, and I'm sorry for those messy bits :disappointed: 
I fixed the container/component pattern issue, I also added the center object to the params that I'm updating, but I didn't notice any performance boost, did I miss something?
I also added the back button to the data-error case, check the screenshot below:
Screenshot from 2019-07-29 16 41 38

fannycc:
Hey! Back button and share icon look good, for designing the rest of the section (text margins and gif find asset here) -> https://www.pivotaltracker.com/n/projects/2321903/stories/167093645
and design here -> https://projects.invisionapp.com/d/main#/console/17484384/376278599/preview

kfilip94:
@fannycc Thanks, I didn't see this task before, I'll add the styles in this PR :+1:

kfilip94:
@fannycc Done :heavy_check_mark: check this gif below:
fhr28-q09cb

src/components/landscape-sidebar/geo-description-widget/geo-description-widget.js
```diff
@@ -1,7 +1,19 @@
...
 import mapStateToProps from './geo-description-widget-selectors';
+
 // this forces the registration of redux module and sagas
 import 'redux_modules/geo-description';
 
-export default connect(mapStateToProps, null)(Component);
+const GeoDescriptionWidget = props => {
+  const handleBackClick = () => {
+    const { view } = props;
+    const target = { center: view.center, zoom: 7 }
```
weberjavi: I've been testing the `goTo` function and it seems that the delay on the movement on the camera is due to the way we are passing the params to the function. It works fine by setting the `position`. Something like this:
```js
const handleBackClick = () => {
    const { view } = props;
    const { center } = view;
    const target = {
      position: {
        x: center.longitude,
        y: center.latitude,
        z: 1000000
      },
      heading: 0,
      tilt: 0
    }
    view.goTo(target)
  };
```

kfilip94:
@weberjavi Thank you, finally fixed! :man_mechanic:

PR #85, opened 2019-08-01
Add switcher

aabdaab:
PT task: https://www.pivotaltracker.com/story/show/167457163

Add a switcher for switching between experiences on the globe.

As we are using two web scenes, one for each mode, the previous setup of the app with a hard refresh between web scenes was causing poor user experience - switching was slow.

This PR introduces a change in the setup, pulling out a parent component to the top-level (`<Experiences />`) which then lazy-loads two experiences: feature and expert.
Few effects of that approach:
- The first load time is not affected (in opposed to loading two web scenes at once in order to smooth the switching UX)
- The second web scene is lazy-loaded, so it is ready to be rendered on the screen, check the network requests comparison between develop and half-earth-demo for instance:
Develop with lazy-loading:
image

Half-Earth-Demo without lazy-loading:
image

- Every next switch is instantaneous as it is already loaded and cached
- Url structure has been preserved
- Query params are preserved on switching


src/components/switcher/switcher-styles.module.scss
```diff
@@ -0,0 +1,36 @@
...
+  top: 0%;
+  left: calc(50% - #{$half-width});
+  width: 340px;
+  height: 38px;
+  clip-path: polygon(0 0, 100% 0, 88% 100%, 14% 100%);
+  box-shadow: 0 2px 10px 0 rgba(23,27,48,0.15);
+  background-color: $elephant;
+  color: $white;
+  letter-spacing: 1.5px;
+  font-size: $font-size-xxs;
```
kfilip94: Please use here `%subtitle`, one of the placeholder Selectors from typography-extends.scss

src/pages/experiences/experiences-component.jsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import cx from 'classnames';
+import loadable from '@loadable/component'
+import styles from './experiences-styles.module';
+
+const ExpertMode = loadable(() => import('./data-globe'));
+const FeaturedMode = loadable(() => import('./featured-globe'));
```
weberjavi: For the loadable pattern to work we need to add a condition on the `render` of the `loadable component`. The way I was envisioning  was to add a watcher to the `onLoad` function of both `Globes` that would add a param to the URL whenever the first `Globe` was rendered on the screen so only then we would be `mounting` the second `Globe` and hence start the `fetch` of its dependencies.
Imagine that scenario for the `data-globe`:

Set some initial state for the mount params

`data-globe-initial-state.js`

```js
export default {
// ... more initial state attributes 
  globe: {
    mountFeaturedGlobe: false,
    mountExpertGlobe: true
  }
}
```

Update the param the first time the globe renders on the screen

`data-globe.js`

```js
const handleMapLoad = (view, setDataGlobeSettings) => {
 // .. more stuff to do on load
  loadModules(["esri/core/watchUtils"]).then(([watchUtils]) => {
    watchUtils.whenNotOnce(view, 'updating', () => { setDataGlobeSettings({ mountFeaturedGlobe: true });})
  })
}
```

Check whenever the URL param changes and load the component when it is set to `true`

`experiences-component.jsx`

```jsx
<>
  <div className={cx({ [styles.hidden]: !isExpertMode })}>
    {mountExpertGlobe && <ExpertMode handleSwitch={() => { preserveParams(switchToFeature) }} />}
  </div>
  <div className={cx({ [styles.hidden]: isExpertMode })}>
    {mountFeaturedGlobe && <FeaturedMode handleSwitch={() => { preserveParams(switchToExpert) }} />}
  </div>
</>
```

src/components/switcher/switcher-styles.module.scss
aabdaab: 👍

aabdaab:
@weberjavi I added the mounting params, ready again ☝️

PR #86, opened 2019-08-07
Feature/preload assets 

weberjavi:
This PR was intended to built the layout for the first view on `Featured maps` but instead it turned out to be more an architecture one.

We had decided to move away from pages (one for each globe) to `mounting` both globes on the same page and dynamically displaying just one of them depending on the `path` the user was joining. This improved the the time it took on switching between experiences but we had not anticipated that although the second scene is hidden it stills receives the props (because is `mounted`) hence performing all the computations. Beyond being a performance leak it turned out to be a mayor bug since we are relying on the `view` prop of the globe to do some `extent` calculation and when this is not rendered there is no `view` 🙈 

So I reverted things back to the page set up. 
Have added some [performance improvements](https://www.keycdn.com/blog/resource-hints) by using `link` `preload` and `preconnect` (thanks @sorodrigo fro your wise advice). So now we are fetching some resources to the data experience while the featured experience is _idle_ this decreases a bit the first load of the second experience and forces caching. Some further improvements could be made though (we should update the `cache-control` for some of the assets) and try to also improve our bundle size.

Further down the road would be clever to use the `Scene` component of the `react-arcgis` library instead of the `Webscene` one. That way we would be able to load layers an esri modules as needed instead of downloading all the dependencies on the first load.

Before all this (when the PR was still a featured maps thing) had set in place the Contentful functionality to fetch the featured maps data (title, description and picture). That data could be updated under the `Featured maps` `content type` on Contentful.

Have fun @simaob :)

Kapture 2019-08-07 at 9 47 31


simaob:
thanks :P

public/index.html
```diff
@@ -7,10 +7,25 @@
       name="viewport"
       content="width=device-width, initial-scale=1, shrink-to-fit=no"
     />
+    <!-- Preconect to usual resources-->
+    <!-- Reference: https://www.keycdn.com/blog/resource-hints#preconnect -->
+    <link href="https://localhost:3000" rel="preconnect" crossorigin>
```
simaob: is this for production use or only for local use?

weberjavi: This is for `development` testing. Probably we would need to update that to `half-earth.vizzuality.com` before adding it to prod.

sorodrigo: Preferably this would be an env variable, so we never forget to update it.

simaob: Yeah, that sounds like a good way of doing it.

src/constants/preload-urls.js
```diff
@@ -0,0 +1,24 @@
+export const DATA_GLOBE_URLS = [
```
sorodrigo: this is nice, I think you should also have the preconnect links here.

public/index.html
weberjavi: Thanks. I will add this `preconnect` resources dynamically (like the `preload` ones) as @sorodrigo says in the other comment.

sorodrigo: Sorry I mightve suggested something that's incorrect. The preconnect links should probably be server rendered, how ever if possible it would be nice to add them from a constants list of some sort, adding the API_URL as an env var.

In trase we're using a webpack plugin, perhaps you can make it work with CRA?
https://github.com/joaner/html-webpack-preconnect-plugin

If it's not possible to do it, just using a env var and writing it down in the index would be good enough ✌️

simaob:
I didn't look super thoroughly at the code as I'm not that proficient in react, but overall it looked tidy =)

weberjavi:
@simaob also found [this legend bug](https://www.pivotaltracker.com/n/projects/2321903/stories/167751596) (on Safari and Firefox) this morning.

Do you want me to address it on this PR or open a new one?

simaob:
ah, I missed that. I saw it on Chrome. So maybe it's an all rounder. Feel free to fix it in a different PR.

public/index.html
simaob: Is there an immediate action regarding this in this PR? otherwise it would be good to go.

weberjavi:
Merging that one with the fixed legend

PR #87, opened 2019-08-08
Feature/default experience layout

weberjavi:
This PR sets the first view of the featured maps experience. By default it has 'Best places in the Biosphere' selected (as is the only featured points we currently have).
Almost two pivotal tickets are addressed:
https://www.pivotaltracker.com/story/show/167590603
https://www.pivotaltracker.com/story/show/167591610

I have set some constrains to the `scene` regarding zoom levels (now we can see the earth slightly smaller on the lowest levels and we can go closer than 10000 meters).

Feel free to check the functionality on the demo site :)

P.S: Note that markers styles and featured map list styles are yet to be implemented in other PRs


weberjavi:
@simaob there's a specific [task about the markers](https://www.pivotaltracker.com/story/show/167591640). We may be creating a custom ones so probably worth handle all marker relates stuff on that PR.

weberjavi:
So... any one gives me approval on this lonely GitHub Friday?
@simaob, @fannycc ?

PR #88, opened 2019-08-19
Feature/featured place detail card

weberjavi:
This PR adds the `featured place card` component to the map and all the logic behind it.
[InVision](https://projects.invisionapp.com/d/main#/console/17484384/374973451/preview)
[Pivotal](https://www.pivotaltracker.com/story/show/167591713)
Whenever a featured place marker is clicked the data for that place is displayed on the card. Clicking on `explore this area` will trigger landscape mode (the elements of the landscape mode are yet to be added, more on [this Pivotal task](https://www.pivotaltracker.com/story/show/167590243))

Kapture 2019-08-19 at 17 20 57

There are some things that need some refinement and experimentation:
- Camera perspective is not the same as in the designs (though I like the current state, what do you think @fannycc  :), further development should be done to accomplish the same view :(
- Scrolling of the card should behave differently, take this PR as the first iteration on this feature. Probably it will need some refinement session with some handsome designer (@fannycc , @faustoperez 😅 )
- All layers are being displayed right now. Arcgis layer manager component will eventually fix that.


aabdaab:
Hey @weberjavi, did you maybe forget to push some commits? For me, the arrows are not working nor I can't see the `Explore this area` trigger 🤔

aabdaab:
Never mind, I had this branch already on my computer and didn't fetch the changes

aabdaab:
Could we reset the scroll when switching between places? 
snapshot-giff

aabdaab:
Really smooth animation of the place card on full screen and on the point click! Maybe we could also add it on clicking the back button?

src/components/featured-place-card/featured-place-card-styles.module.scss
```diff
@@ -0,0 +1,162 @@
...
+    width: 100px;
+    padding: 0 10px;
+    height: 100%;
+    background-color: $white;
+    display: flex;
+    flex-direction: row;
+    justify-content: space-around;
+    align-items: center;
+
+    .arrowWrapper {
```
aabdaab: I don't think we need to nest it; we're not overriding it

weberjavi:
Hey @aabdaab thanks for the review. Gonna address this issues, will let you know when they are fixed 🙌

weberjavi:
> Really smooth animation of the place card on full screen and on the point click! Maybe we could also add it on clicking the back button?

Have been trying to trigger the animation on component `unmount` with the `useTransition` hook from `react-spring` but it is not working. So finally I've opted for avoid lay loading the component so that way it transitions out smoothly instead of desappearing when ùnmount`

src/hooks/featured-place-view-manager-hooks.js
```diff
@@ -0,0 +1,32 @@
+import { useEffect, useState } from 'react';
+
+export function useFeaturedPlaceViewCameraChange(view, selectedFeaturedPlace, featuredPlacesLayer) {
+const [coords, setCoords] = useState(null);
+useEffect(() => {
+  if (selectedFeaturedPlace && featuredPlacesLayer) {
+    const query = featuredPlacesLayer.createQuery();
+    query.where = `nam_slg = '${selectedFeaturedPlace}'`
+    featuredPlacesLayer.queryFeatures(query).then(result => {
+      const { lon, lat } = result.features[0].attributes;
```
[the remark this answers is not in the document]
weberjavi: Not sure if `get` is safer in this situation, if some property on the path is undefined an error will trigger when trying to destructure. I would maybe avoid importing `lodash` in this case. What do you think @aabdaab ?
 From `lodash` REPL:
image

src/components/featured-place-card/featured-place-card-styles.module.scss
weberjavi: Not sure what you mean @aabdaab can you clarify this comment?

src/styles/typography-extends.scss
```diff
@@ -2,8 +2,9 @@
 
 // Big titles, HE-Meter page
 %display1 {
-  font-family: $font-family-1;
+  font-family: $font-family-2;
```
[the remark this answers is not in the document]
weberjavi: ⚡️

src/components/featured-place-card/featured-place-card-styles.module.scss
aabdaab: so, I meant that we don't have to nest `arrowWrapper` class inside `placesNavigator` (as we are not overriding this class)

weberjavi: Ah ok. Well is just a matter of taste, I mean if you look closer the whole `scss` file I use to nest the styles to mimic the `html` structure, to have a grasp of the node tree structure when writing the styles.
For example this structure on the styles
```scss
.navigation {
  .backToMap {}
  .placesNavigation {}
}
```
resembles this one from the markup
```html
<nav className={styles.navigation}>
  <div className={styles.backToMap}/>
  <div className={styles.placesNavigation}/>
</nav>
```
Does it makes sense?

aabdaab: I'm not sure about that approach. Generally, nesting css classes means that you're overriding something, i.e. my first thought when I see 
```
.navigation {
  .backToMap {}
}
```
is 'ok, so where is the original `backToMap` class and what are you overriding?'
Also, have a look at this [article](http://thesassway.com/intermediate/avoid-nested-selectors-for-more-modular-css), it talks about a more modular css (in our case it is already modular as we've got css-modules but it's more of a matter of convention at this point)

src/hooks/featured-place-view-manager-hooks.js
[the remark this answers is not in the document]
aabdaab: in the case, you showed, yes, destructuring would throw an error, same as here:
```
const { lon, lat } = result.features[0].attributes;
```
To take advantage of `get` we would have to change it to:
```
const lon = _.get(result, 'features[0].attributes.lon');
const lat = _.get(result, 'features[0].attributes.lat');
```

image

I don't have a strong opinion about this although I'm a bit worried not checking for undefined in deeply nested objects (i think it's better to set an undefined just in case than throw an error to a user)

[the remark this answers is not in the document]
weberjavi: Agree 👍

src/components/featured-place-card/featured-place-card-styles.module.scss
weberjavi: Well when I see nested classes I think on namespacing :)
But yeah probably what we are talking about here is about setting some sort of `best practice` for the project. As we are already being modular and namespacing by default with `css-modules` we are avoiding a lot of the problems that will arise in other context. But we might want to define some convention for the styles.
Do you think having some conversation about this would be beneficial? @aabdaab , @kfilip94 ?

weberjavi:
☝️ ready again

src/components/featured-place-card/featured-place-card-styles.module.scss
kfilip94: Well, we don't need to nest things there, but I also understand that it can help to reflect the HTML structure. I don't have a strong opinion about one or another approach, I used both sometimes. If you want we can pick the one approach but to me, it is not something super important. Just use the approach that you prefer. :wink:

kfilip94: Well, unless you want to nest more than two levels what is redundant for sure and can make the code less readable.

fannycc:
> Overall it's looking good, I left some comments about minor changes. We should also think about the `About Half-Earth Map` button hanging over the place card on smaller screens (also when you hover over it, it disappears as it is white, the same as the background color of the card) but this is more of a question for @fannycc and to address probably when we start to implement responsive designs.

The button for about half-earth map shouldn't appear on that view (as the Invision shows) we are focusing in the card the user selected in that status, other elements are noise at that point.

weberjavi:
@aabdaab , @kfilip94 updated this branch with the proper icon on `explore this area` button and removing `about` button when a `featured place` is selected. Do you think we can follow up the conversation about `css` naming conventions and code structure in another place (maybe set a meeting to decide on which convention to use and create a document to make it explicit)?
If you think there's some further action to take on this PR feel free to request those changes.

weberjavi:
btw `demo` site is up to date with this branch in case anyone wants to take a look :)

src/components/featured-place-card/featured-place-card-styles.module.scss
aabdaab: Let's have a meeting next week -- it's not that something is breaking because of this

aabdaab:
> Could we reset the scroll when switching between places?
> snapshot-giff

did you have a look at the scroll resetting? Maybe we could still address it in this PR if it's not a major change

aabdaab:
> > Could we reset the scroll when switching between places?
> > snapshot-giff
> 
> did you have a look at the scroll resetting? Maybe we could still address it in this PR if it's not a major change

- I see you fixed it! I had it cached but then on incognito I see it's working. Thank you 🙌

weberjavi:
🙇

PR #89, opened 2019-08-20
All maps screen

aabdaab:
This PR adds a design implementation of All Maps screen (without categories for now).
PT task: https://www.pivotaltracker.com/story/show/167591835
Design: https://projects.invisionapp.com/d/main#/console/17484384/377925959/preview

* Adds scrollable cards container
* On card hover show related points on the map (via setting slug in url)

screencast 2019-08-20 08-06-42

### DIFF
--- a/src/components/featured-maps-list/featured-maps-list-component.jsx
+++ b/src/components/featured-maps-list/featured-maps-list-component.jsx
@@ -40,7 +40,7 @@ const FeaturedMapsListComponent = ({
     <animated.div className={cx(className, styles.cardsContainer)} style={slide}>
       <div className={styles.scrollable}>
         {featuredMapsList.map(featuredMap => (
-          <div className={styles.card}
+          <div className={cx(styles.card, { [styles.border]: cardOpen.slug === featuredMap.slug })}
             onClick={() =>  handleClick(featuredMap)}
             onMouseOver={() => mouseOver(featuredMap)}
             onMouseOut={mouseOut}
--- a/src/components/featured-maps-list/featured-maps-list-component.jsx
+++ b/src/components/featured-maps-list/featured-maps-list-component.jsx
@@ -1,6 +1,7 @@
 import React, { useState } from 'react';
 import cx from 'classnames';
 import { useSpring, animated } from 'react-spring';
+import { ReactComponent as ArrowExpandIcon } from 'icons/arrow_expand.svg';
 import styles from './featured-maps-list-styles.module';
 
 const FeaturedMapsListComponent = ({ 
@@ -10,7 +11,9 @@ const FeaturedMapsListComponent = ({
   isLandscapeMode,
   isFullscreenActive,
   handleFeaturedMapClick,
-  handleFeatureMapHover
+  handleFeatureMapHover,
+  handle,
+  view
 }) => {
   const isOpen = selectedSidebar === 'featuredMapsList';
   const slide = useSpring({
@@ -19,31 +22,33 @@ const FeaturedMapsListComponent = ({
     delay: isOpen && !isLandscapeMode && !isFullscreenActive ? 400 : 0
   })
 
-  const defaultCardOpen = featuredMapsList[0];
-  const [cardSelected, setCardSelected] = useState(false);
-  const [cardOpen, setCardOpen] = useState(defaultCardOpen);
+  const [cardOpen, setCardOpen] = useState(null);
 
+  const resetCardOpen = () => setCardOpen(null);
   const mouseOver = (card) => { setCardOpen(card); handleFeatureMapHover(card.slug) };
-  const mouseOut = () => {
-    if(!cardSelected) {
-      setCardOpen(defaultCardOpen);
-      handleFeatureMapHover(defaultCardOpen.slug);
-    }
-  };
 
   const handleClick = (card) => {
-    setCardSelected(true);
+    resetCardOpen();
     handleFeaturedMapClick(card.slug);
+    if (handle) {
+      handle.remove();
+    }
   };
 
   return (
     <animated.div className={cx(className, styles.cardsContainer)} style={slide}>
+      {featuredMapsList && <button
+        className={styles.button}
+        onClick={() => handleClick(featuredMapsList[0])}
+      >
+        <ArrowExpandIcon className={styles.icon} />
+        <span className={styles.backButton}>BACK</span>
+      </button>}
       <div className={styles.scrollable}>
-        {featuredMapsList.map(featuredMap => (
-          <div className={cx(styles.card, { [styles.border]: cardOpen.slug === featuredMap.slug })}
+        {featuredMapsList && featuredMapsList.map(featuredMap => (
+          <div className={cx(styles.card, { [styles.border]: cardOpen && cardOpen.slug === featuredMap.slug })}
             onClick={() =>  handleClick(featuredMap)}
             onMouseOver={() => mouseOver(featuredMap)}
-            onMouseOut={mouseOut}
           >
             <section
               className={styles.titleSection}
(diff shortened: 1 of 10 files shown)


weberjavi:
☝️ checking ☝️

src/components/featured-maps-list/featured-maps-list-component.jsx
```diff
@@ -1,25 +1,62 @@
...
     marginLeft: isOpen && !isLandscapeMode && !isFullscreenActive ? 0 : -400,
     delay: isOpen && !isLandscapeMode && !isFullscreenActive ? 400 : 0
   })
+
+  const defaultCardOpen = featuredMapsList[0];
+  const [cardSelected, setCardSelected] = useState(false);
+  const [cardOpen, setCardOpen] = useState(defaultCardOpen);
+
+  const mouseOver = (card) => { setCardOpen(card); handleFeatureMapHover(card.slug) };
+  const mouseOut = () => {
```
weberjavi: I think I would get rid of the `mouseOut` functionality. As the list items `height` is changing whenever are hovered it could drive the interface to undesired behaviours. It also allows the user to move the mouse with out losing the selection (if not the user is forced to have the cursor over the list item to read the text.

__With `mouseOut`__

Kapture 2019-08-20 at 11 11 38
)

__Without `mouseOut`__

Kapture 2019-08-20 at 11 16 17

By the way I think this list item changing `height` is not the best UX solution. Maybe a fixed `height` with the text entering from above would be smoother or maybe just adding some animation/transition to the growing `height`. But that's a question for @fannycc :)

src/components/featured-maps-list/featured-maps-list-styles.module.scss
```diff
@@ -0,0 +1,53 @@
...
+
+.cardsContainer {
+  display: flex;
+  flex-direction: column;
+  width: $sidebar-width;
+  border-radius: 2px;
+  transition: all 0.2s ease-out;
+}
+
+.scrollable {
```
weberjavi: I think you can get rid of this `scrollable` `class` and just add `@extend %verticalScrollBar;` to the `cardsContainer`.

src/components/featured-maps-list/featured-maps-list-component.jsx
```diff
@@ -1,25 +1,62 @@
...
+  handleFeatureMapHover
 }) => {
   const isOpen = selectedSidebar === 'featuredMapsList';
   const slide = useSpring({
     from: { marginLeft: -400 },
     marginLeft: isOpen && !isLandscapeMode && !isFullscreenActive ? 0 : -400,
     delay: isOpen && !isLandscapeMode && !isFullscreenActive ? 400 : 0
   })
+
+  const defaultCardOpen = featuredMapsList[0];
```
weberjavi: If we avoid the `mouseOut` functionality (as recommended on other comment) we will not need this `defaultCardOpen` as the card open will be the `featuredMap` selected (that also would make sense when reaching the `featuredMapsList` from a `featuredMap` as this will be the opened list item).
BTW on the InVision designs the default state is with all the items closed (not sure if we should remove this `defaultCardOpen` anyway).

fannycc:
This view "all maps" needs to be very different from the previous screen. All elements but the cards and the 🌎 should disappear. The globe here is just used as an example of what is going to appear on top of it with every map. Users SHOULDN'T be able to interact with it. Ideally it spins to show all the markers non top and it needs to be placed on the right side of the screen. The desirable transition is that, after the elements of the UI disappear, the globe can be placed in the right position so the cards appear on its left. Last time I talked with @weberjavi I told them I'm aware that moving the globe around the interface might be tricky, if we try it and we don't succeed, we can discuss other options.

aabdaab:
@fannycc 
- The globe here is just used as an example of what is going to appear on top of it with every map. Users SHOULDN'T be able to interact with it. --> but it should be visible? just not interactive?

fannycc:
> @fannycc
> 
> * The globe here is just used as an example of what is going to appear on top of it with every map. Users SHOULDN'T be able to interact with it. --> but it should be visible? just not interactive?

yup! the globe should be there, is it possible to block interaction with it and use it as a placeholder? it would be the same case than the "about half-earth project" Chris wants the conservation areas appear on top of it but the user doesn't interact with it. We use the globe as a placeholder in both cases and it should have the same basemap so the user can build mental models between all the experience across the platform and we are consistent. Happy to talk about this in a call

weberjavi:
@aabdaab maybe this is useful in order to disable events on the globe:
https://developers.arcgis.com/javascript/latest/sample-code/view-disable-zoom/index.html

src/components/featured-maps-list/featured-maps-list-component.jsx
aabdaab: @fannycc what would be your take on that? ☝️

aabdaab: Yes, I definitely like better your approach @weberjavi, thanks! I removed the `mouseOut` and it works smoother

fannycc: hey! so...
1. **about the without mouse0ut approach**, just wondering if it's intuitive that I need to click on it to select that card when the status is already open...also how can I collapse it again? I thought there wouldn't be a big problem that the user needed to leave the mouse on top of the card to get the preview, after all there isn't any other available interaction in the interface than the cards. I'm just a bit a concern of adding this feature which is not consistent with the rest of our hovers. There is a bug in the first option "with mouse0ut" it seems that it gets crazy...I would be happy to see it working with the preview just in hover and then play around and show it to others. 

2. And about the height of the card. If it feels jumpy or weird we can try to fix the same height for the cards...bearing in mind that we have different lengths for the content and some might have a lot of space. Happy to try it out tho! :) then we should fix the longest example and duplicate this behaviour to the map with the selected card. All the cards should have the same height no matter the interface where they appear.

src/components/featured-maps-list/featured-maps-list-styles.module.scss
aabdaab: fun fact: without the additional `.scrollable` class (moving the scroll to cardsContainer) the blur effect stops working 🤷‍♀ so I'll extend the verticalScrollbar in the scrollable class

aabdaab:
Ok, this is ready again!
A couple of things have been added:
- `pointer-events: none` to the globe to disable its interactivity on All maps screen
- All the elements but the `About...` are hidden
- Globe is slightly rotating when on All maps screen
- Scrollbar is styled (just duplicate this code:
```
{featuredMapsList && featuredMapsList.map(featuredMap => (
          <div className={cx(styles.card, { [styles.border]: cardOpen && cardOpen.slug === featuredMap.slug })}
            onClick={() =>  handleClick(featuredMap)}
            onMouseOver={() => mouseOver(featuredMap)}
          >
            <section
              className={styles.titleSection}
              style={ {backgroundImage: `url(${featuredMap.image})`}}
            >
              <h2 className={styles.title}>{featuredMap.title}</h2>
            </section>
            {cardOpen && cardOpen.slug === featuredMap.slug && <section className={styles.descriptionSection}>
              <p className={styles.description}>{featuredMap.description}</p>
            </section>}
          </div>
        ))}
```
to see it
- Back button added (it's taking the first element of the array of featured map list)
- I left the cards height untouched - without the `mouse0ut` it's not that jumpy and to me it seems ok - I'd say let's deploy to demo to have @fannycc play with it and then decide
- The padding of the globe hasn't been implemented as unfortunately it cannot be smoothly animated with the functions that we know of. I'm gonna add a PT task to research it further.

weberjavi:
☝️ checking

src/pages/featured-globe/featured-globe-component.jsx
```diff
@@ -41,51 +43,71 @@ const FeaturedGlobeComponent = ({
...
   mouseMoveCallbacksArray
  }) => {
+  const isMapsList = selectedSidebar === 'featuredMapsList';
+  const [handle, setHandle] = useState(null);
+
+  const spinGlobe = (view) => {
+    loadModules(["esri/core/scheduling"]).then(([scheduling]) => {
+      const spinningGlobe = scheduling.addFrameTask({
+        update: function() {
+          const camera = view.camera.clone();
```
weberjavi: Hey @aabdaab this is quite demanding task for the computer (at least mine computer fan is about to take off :). It gets smoother if we move the camera cloning outside the frame task to avoid cloning the camera on every frame.
🙏

aabdaab: good point, I'll move it outside! thanks a lot for the review 🙏

PR #91, opened 2019-08-21
Feature/landscape sidebar for feature globe

kfilip94:
This PR adds landscape components to the feature mode:
[PIVOTAL](https://www.pivotaltracker.com/story/show/167590243)
- adds landscape sidebar with support of human pressures widget and geo-describer component,
- adds the legend,
- fixes the bug human pressures: having human pressures rasters selected, after closing HP layers by clicking 'x' on the legend, the rasters were still selected on the HP widget in landscape sidebar
- hides the human pressures layers when the view it's not in landscape mode,
msn3p-fd5aa
image

It also changes the structure of arcgis-layer-manager files, I needed to introduce small changes for featured-map, so I divided this component into two different components, one for data globe and the second one for the featured globe.


src/components/data-arcgis-layer-manager/data-arcgis-layer-manager-component.jsx
```diff
@@ -0,0 +1,30 @@
...
+    mapLayer.layers && mapLayer.layers.items && mapLayer.layers.items.forEach(layer => {
+      const setActive = activeLayers && activeLayers.some(activeLayer => activeLayer.title === layer.title);
+      setOpacity(layer, activeLayers);
+      layer.visible = !!setActive;
+      mapLayer.visible = !!setActive;
+    })
+    setOpacity(mapLayer, activeLayers);
+
+    // Set group layers as visible to allow sublayers to be visible
+    if (mapLayer.type === "group") {
```
weberjavi: We also need this statement on the `featuredGlobe` (if we do not have it we are not able to see the `LabelsLayer` as it is a group layer).
It also makes our `initial-state` `activeLayers` to work properly (that happens when we set `mapLayer.visible = !!setActive;`) otherwise the layers will get their initial state from the Arcgis Online `scene` it self (all layers are visible by default there).

src/components/featured-arcgis-layer-manager/featured-arcgis-layer-manager-component.jsx
```diff
@@ -0,0 +1,29 @@
+import { LAND_HUMAN_PRESSURES_IMAGE_LAYER } from 'constants/layers-slugs';
+import { setLayerOrder, setOpacity } from 'utils/arcgis-layer-manager-utils';
+
+const ArcgisLayerManager = ({ map, activeLayers, isLandscapeMode }) => {
```
weberjavi: I think this is a good place to extend the functionality of the `ArcgisLayerManager` component by creating a param where we can inject some function call inside the layers loop.
That way instead of duplicating the component forcing us to maintain both of them (and eventually more if we end up creating more scenes) we would be extending the functionality as needed.
I can imagine something like:

on `featured-arcgis-layer-manager-component.jsx`:

```jsx
import LayerManager from `arcgis-layer-manager`;
const FeaturedGlobeLayerManager = ({map, activeLayers, customFunctions}) => {
  return <LayerManager map={map} activeLayers={activeLayers} customFunctions={customFunctions} />
}
```
on `featured-arcgis-layer-manager.js`:

```js
const hideHumanPressuresOnLandscape = (layer, setActive) => {
if(layer.title === LAND_HUMAN_PRESSURES_IMAGE_LAYER) {
    layer.visible = isLandscapeMode && setActive;
 }
}
const customFuntions = [
  hideHumanPressuresOnLandscape
]
```

on `arcgis-layer-manager-component`:

```js
const ArcgisLayerManager = ({ map, activeLayers, customFunctions }) => {
  layers.items.forEach(mapLayer => {
  // ...layer manager stuff
  customfunctions.forEach(fn => fn(mapLayer))
  })
}
```
With something along those lines we would have a place to inject custom functionality or conditionals dependent on the scene needs but would have only one component with this layer toggling functionality that is paramount for the app working and is kinda complex and ugly (and that's another story that we might want to address in the future :)

src/pages/featured-globe/featured-globe-component.jsx
```diff
@@ -45,6 +59,8 @@ const FeaturedGlobeComponent = ({
         { DATA_GLOBE_URLS.map(url => <link key="url" rel="preload" crossorigin href={url} as="script" />)}
       </Helmet>
       <Globe sceneId={SCENE_ID} sceneSettings={sceneSettings} onLoad={onLoad} loadElement={<Spinner spinnerWithOverlay />}>
+        {isGlobeUpdating && <Spinner floating />}
+        <ArcgisLayerManager activeLayers={activeLayers} isLandscapeMode={isLandscapeMode} isFeatureMode/>
```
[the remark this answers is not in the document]
kfilip94: I forgot to remove that, thanks :+1:

src/pages/featured-globe/featured-globe-initial-state.js
```diff
@@ -1,12 +1,25 @@
...
+  GRID_LAYER
+} from 'constants/layers-slugs';
+
 export default {
   globe: {
-    activeLayers: [],
+    activeLayers: [
+      { title: BIODIVERSITY_FACETS_LAYER }, // Biodiversity Facets (new grid)
+      { title: GRID_LAYER }, // This is the layer used to paint aggregated grid cells
+      { title: LABELS_LAYER_GROUP }, // This is the layer used to paint aggregated grid cells
```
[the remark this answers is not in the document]
kfilip94: Done :+1:

src/components/featured-arcgis-layer-manager/featured-arcgis-layer-manager-component.jsx
kfilip94: That looks much better, changes are in the repo now. Also, I removed `featured-arcgis-layer-manager` and `data-arcgis-layer-manager` components, and I moved `customFunctions` to the `featured-globe.js` container because having these two separate components seems to be redundant now. Thank you for this hint, I like it better! :1st_place_medal:  :sunglasses:

src/utils/arcgis-layer-manager-utils.js
```diff
@@ -0,0 +1,24 @@
+import { includes } from 'lodash';
+import { LEGEND_FREE_LAYERS } from 'constants/layers-groups';
+
+export const setLayerOrder = (activeLayers, map) => {
```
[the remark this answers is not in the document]
kfilip94: :v:

src/components/data-arcgis-layer-manager/data-arcgis-layer-manager-component.jsx
kfilip94: Done :ok_hand:

kfilip94:
Thanks for your review Javi! :+1: ready again, check the answers above :point_up_2:

weberjavi:
☝️ checking

PR #93, opened 2019-08-26
Update human pressures widget

kfilip94:
This PR updates the human-pressure-widget in landscape-sidebar by introducing bar visualization of selected pressures.
[PIVOTAL](https://www.pivotaltracker.com/story/show/167900080) | [INVISION](https://projects.invisionapp.com/d/main#/console/17484384/365514646/inspect)
ll2ri-jffqi


weberjavi:
☝️ checking

weberjavi:
@kfilip94 I see this bug is not happening on your `gif` let me know if you can not reproduce it.

kfilip94:
@weberjavi Thanks for your review! :+1:  Sadly, I cannot reproduce this blur bug. :( 
Which version of Chrome are you using?
I can take a look at the human-pressures bug, it's weird that it happens only after scenes switch, any hints what can cause this kind of problem?

weberjavi:
Using latest Chrome version.
image

It is also happening on safari (also when the checkbox is selected):
image

kfilip94:
Well, I have no idea then. @aabdaab could you check it on your machine, when you have a moment? thanks

kfilip94:
@weberjavi Fixed & ready again. You have to have at least one child with `position: fixed` inside the container with blur background, and when all checkboxes are not selected, we don't have any tooltips in DOM what means we don't have any elements with `position: fixed` as well. I don't know why it works but it does, probably there is a more logical solution to this issue but I don't have a MacBook machine to dig deeper into that. Please check it on your machine :pray: 

Thanks for help @aabdaab! :+1:

PR #96, opened 2019-08-28
Bug/landscape mode camera

weberjavi:
This small PR is a quick fix to avoid this bugs:
- When going back to featured place detail card from `landscapeMode` we where using `position.z` to set the meters above the surface to drive the changes on the camera, this was not behaving as expected in some situations (depending on the screen size and latitude of the place). I've changed that to use the `zoom` level instead, that is the param that really drives this camera update. 
- Also added `isLandscapeMode` prop as a dependency in the `useFeaturedPlaceViewCameraChange` hook to avoid messing with the landscape camera settings.

PS: Eventually we might want to 'centralise' all this camera functionality in some sort of `<CameraManager />` component, what do you think?


src/components/landscape-sidebar/geo-description-widget/geo-description-widget.js
```diff
@@ -9,17 +9,7 @@ import 'redux_modules/geo-description';
 const GeoDescriptionWidget = props => {
   const handleBackClick = () => {
     const { view } = props;
-    const { center } = view;
-    const target = {
-      position: {
-        x: center.longitude,
```
kfilip94: By removing `target` object with position inside, we are messing the performance of the animation. Now we have 2 seconds delay between the camera starts moving.
This is the exact same issue as we were trying to resolve in my previous PR, check your [comment here](https://github.com/Vizzuality/half-earth-v3/pull/84#discussion_r308611055)

kfilip94:
Hey @weberjavi, could you please describe me better the issue or attach the gif with the bug? Because I'm struggling a bit with reproducing this :bug:  Thanks :)

weberjavi:
So sometimes the `back` button does not take us out landscape mode (because of the `position.z` value). Let me check if I can fix the `delay`
Kapture 2019-08-28 at 12 56 43

weberjavi:
Ok so have updated the `goTo` invocation using `scale` instead of `zoom`. The speed of the interaction changes from click to click, it is not a super stable interaction, I think there are a lot of things happening in the thread at the same time. We would need to face this performance issues as soon as possible.
The `camera` changes are working fine now (when we want to exit `landscapeMode` we do and the `camera` is properly positioned).

Kapture 2019-08-28 at 17 18 14

weberjavi:
@kfilip94 , @aabdaab ready again ☝️

kfilip94:
@weberjavi thanks, I think we can merge it. :+1: 
I thought that passing `position` object will solve the delay problem, but after playing with this I'm not sure.

PR #97, opened 2019-08-28
Feature/caching strategies

weberjavi:
This PR creates a custom Service Worker. 
The service worker gets added to the `bundled` code after the `build` process. This step is triggered by the `build-service-worker` script and the config of the worker is defined on the `service-worker-build.js` file. That generates a `custom-service-worker.js` file that follows the structure set in place on the `service-worker-template.js`.
This last file is the one that should be modified in order to update the service worker. It has a version number that should be updated on every modification.
The files cached on `installing` phase ([service worker lifecycle for reference](https://developers.google.com/web/fundamentals/primers/service-workers/lifecycle)) currently are the app own `js`, `css`, `svg` and `html` assets.
Once the service worker is active it will start intercepting the browser requests and it will currently be listening to cache images and fonts requests (from the app url). So those assets will not be served from the cache until a reload of the page have taken place.
Eventually some other requests could be intercepted to increase the assets cached.

Have tried to cache ArcGIS Js API requests but the worker in not working as expected.

I've added a new dependency: `workbox-build`

To test it locally run: 
`yarn build`
then open the build folder on a server (I'm using `serve`), do not forget to open it on `localhost:3000` as is one of the urls whitelisted on the Biodiversity Facets layer.
`serve build -l tcp://localhost:3000`
and check the `Network` and `Application` tabs on your browser console (you should see the Service worker there and the cache :)

image


aabdaab:
Also, one thing I noticed (I'm not sure if it's caused by that service worker) is that after testing that PR, I launched climate watch (also on port 3000) and I'm unable to open it 🤔However, on incognito I don't have that issue. Could you check if you're experiencing a similar issue with a different app than half-earth?

weberjavi:
> Also, one thing I noticed (I'm not sure if it's caused by that service worker) is that after testing that PR, I launched climate watch (also on port 3000) and I'm unable to open it 🤔However, on incognito I don't have that issue. Could you check if you're experiencing a similar issue with a different app than half-earth?

Yeah @aabdaab this is most def due to the Service Worker is connected with the URL that created it in the first place, so when testing it locally is `localhost:3000`, so the same origin for both apps, and due to that probably the `index.html` served to CW is not the proper one hence crashing the app.
You would probably need to remove the service worker through the `Application` tab of the browser dev tools. That should not be a problem thought when the apps are on staging or prduction, as the URL origin of the service worker will be different.

weberjavi:
> The caching is definitely working 🙆 I never used service workers before so unfortunately, I don't have any advice on how to advice pitfalls in the future, I guess we'll find out (I was reading [this](https://novemberfive.co/blog/mess-up-service-workers-caching-gotcha#targetText=Mess%20up%20%26%20learn%3A%20service%20worker%20caching%20gotchas&targetText=Service%20workers%20are%20a%20great,load%20times%20on%20slower%20networks. and felt a bit at unease 😅)
> I'd be interested to hear from @sorodrigo if he has time 🙏

Yeah I've also read this blogpost... scary 👻 
The Service worker set up creates a content hash by default for all the requests that are not name hashed by default (e.g webpack chunks contain a hash on their generated names). That is, when the content on the file change the hash is updated and then the file requested again, as its new hash is not the same as the old one. That prevents the Service Worker from 'taking control' 
We could test how it works on the `demo`site in case it introduces any unwanted caching.
Also I've added some name spacing to the cache entries.

weberjavi:
@aabdaab let me know your thoughts about my responses ☝️

aabdaab:
I'm happy to merge it and deploy to the demo page and see it in action before HE day. We could even deploy it today... (given also that this task belongs to the last sprint). So when next time we update the demo we will see how it acts with the changes. How does that sound 🎵?

weberjavi:
🙌 merging then 🙌

PR #99, opened 2019-09-02
Location and zoom widgets are dependant on isNotMapsList prop

aabdaab:
Pass `isNotMapsList` prop to Location and Zoom widgets on data-globe experience. These props were added because of Feature mode requirements (All maps screen).


kfilip94:
Now they are back!
But we have another bug, I'm not sure if this is tied with `isNotMapsList` flag, but after a click on location button, I'm getting this error:
image
Could you take a look at this?

aabdaab:
checking

aabdaab:
check now, "it's working on my machine"

PR #101, opened 2019-09-04
Optimize animations

aabdaab:
This PR removes the `react-spring` dependency and introduces pure CSS animations instead.
That change is aiming to speed up the performance of the app.
There was an issue with animating landscape sidebar when **leaving** the landscape mode because of the unmounting of the component (the node gets removed and there's no node to apply the animations to). So, instead of mounting and unmounting landscape sidebar conditionally:
```
{isLandscapeMode && <LandscapeSidebar
            isLandscapeMode={isLandscapeMode}
            isFullscreenActive={isFullscreenActive}
            handleGlobeUpdating={handleGlobeUpdating}
            activeLayers={activeLayers}
            rasters={rasters}
            setRasters={setRasters}
            setLayerVisibility={setLayerVisibility}
/>
```
we hide it when it's not landscape view (we did it before + the unmounting). The animation has time to be applied now.
As a side note, react-spring has a `native` flag for better performance. It's applied by default (see [here](https://github.com/react-spring/react-spring/issues/681)) so we were already using it, hence the moving to pure CSS animations.


aabdaab:
Thanks Javi, that's a good idea, I'll make a mixin out of it! Also, as I was preparing for a talk for frontismos meeting I read some interesting stuff that I didn't know earlier so I'm gonna test some more things and let you know how it worked out 🙂it's about *critical rendering path*, maybe we can optimize the animations even more

aabdaab:
This is ready again ☝️
After learning about **critical rendering path** I added improvements to the CSS animations. The path goes like this: Styles -> Layout -> Paint -> Composite. Styles are the heaviest for the browser to animate therefore to achieve performant animations we need to focus on changing properties that affect the Composite step and not add the stress to the previous layers like for example Layout. We should avoid using **transitions** in combination with **margins/top/left/bottom/right** props as they have the browser creating layouts each time (each animation frame). However, **transform** properties affect a Composite step. Another trick would be to force the browser to handle the animations on the GPU with `will-change` prop; in our case, it did slow things down a bit (globe uses GPU), so I didn't include it. The biggest perceived improvement is when you turn CPU throttling on and measure the FPS.

weberjavi:
☝️ checking

aabdaab:
Check out the last commit; I added the delay and used `@include...` where it made sense. However, I had to leave `.transformOpacityWithDelay` in the `common-animations.module` as something like this:
```
@include animationFunction();
@include animationFunction($propertyToAnimate: opacity);
```
did not work (they have to override each other). I figured it's not worth the time for now to form a more complicated mixin that would maybe take an array of properties and animate them - whenever we add more animations and it becomes a common use case, we can add a mixin like that. Thanks for your thorough review as always!! Gonna merge that now 🚀

PR #102, opened 2019-09-09
Feature/grid layer refactor

weberjavi:
This PR started with the assumption that the calculation of the aggregated grid cell was one of the bottlenecks on the landscape view mode transition so the [initial idea](https://www.pivotaltracker.com/story/show/168213659) was to offset the grid cell calculation into a web worker. 
While developing found out that that was not the case and that some of the improvements that could be made where related to the way we where handling the `watching` of the events and also a lot of requests could be avoided by removing the `biodiversity facets layer` from the scene and instead of querying the features on the `layerView` just query to the `featureLayer` service using the `extent` geometry.
This PR could only be checked on the data globe (as is there where `biodiversity facets layer` is been removed)
Also updated the code to leverage some of the updates made in the last version of the ArcGIS js API (on the previous version the only allowed `spatialRelationship` for querying a feature layer was `intersects` so further calculations where made once we got the features back from the query).

The initial testing made to set up an environment to use web workers could be checked on [this `branch`](https://github.com/Vizzuality/half-earth-v3/tree/feature/web-worker-arcgis), this code could be useful in the future in case we need to use any worker on the platform.

P.S: The land use widget [bottom statement](https://www.pivotaltracker.com/story/show/168215443) has been removed in this PR.


aabdaab:
checking ☝️

src/components/grid-layer/grid-layer-component.jsx
```diff
@@ -31,74 +39,57 @@ const GridLayer = ({map, view, setGridCellData, setGridCellGeometry, handleGlobe
...
-                  // Add it to the store
-                    setGridCellGeometry(gridCellGeometry)
-                  })
-              }
-              gridCellRef.current = gridCells;
-            })
+  useEffect(() => {
+    if (biodiversityFacetsLayer && gridCellGraphic) {
+      const containedCellsQueryObject = containedQuery(biodiversityFacetsLayer, view.extent);
+      biodiversityFacetsLayer.queryFeatures(containedCellsQueryObject)
```
aabdaab: I think we can extract this bit to a function passing `query` and type of a cell; `singleCell` or `aggregatedCells` as it is almost the same in the if and else body

weberjavi:
@aabdaab sorry this comments went under my radar, will check them now :(

weberjavi:
Not sure about the performance improvement. I can say that the speed of rendering the grid and the display of the data on the land pressures widget has increased, as we are not waiting until the `gridViewLayer` stops `updating` but we do perform the features query once the view is stationary and we have the new extent (that not just triggers the redrawing of the grid cell but also the update of the land pressures widget and the fetching of the geo description). Also we are now performing one or two requests (in the case there's no `contained` grid cells on the `extent` to draw the centre one) instead of the `40` requests we where making to update the viewLayer each time we changed the `view extent` .

**This branch**:
image

**Master branch**:

image

src/components/grid-layer/grid-layer-component.jsx
weberjavi: I've extracted the internal logic after the `query` on the `if/else`  statement to a function. But I'm afraid that the query needs to seat outside of this function as they are nested queries, if we do not get multiple grid cells only then we perform a second query inside the `callback`

src/components/grid-layer/grid-layer-component.jsx
```diff
@@ -31,74 +39,57 @@ const GridLayer = ({map, view, setGridCellData, setGridCellGeometry, handleGlobe
...
   }, [])
 
-  // store grid view layer
   useEffect(() => {
-    const { layers } = map;
-      const gridLayer = layers.items.find(l => l.title === BIODIVERSITY_FACETS_LAYER);
-      view.whenLayerView(gridLayer).then(function(layerView) {
-        setGridViewLayer(layerView);
-      })
+    loadModules(['esri/layers/FeatureLayer']).then(([FeatureLayer]) => {
```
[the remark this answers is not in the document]
weberjavi: I hope that eventually this `request` is handled from the `Service Worker` and cached after the first time so subsequent requests to that layer are served from the `cache`. But as we are unmounting the component when we get out from landscape view we are losing the reference. We could add this to the store and check if it is there before fetching again. Not really keen on adding this to the store (just to leave it as terse as possible)but maybe we can do it in another PR, what do you think?

weberjavi:
@aabdaab thanks for the review!!
Please check the code update and the comments ☝️

src/components/grid-layer/grid-layer-component.jsx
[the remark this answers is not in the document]
aabdaab: I see! I'm happy to create a PT task with that idea (moving it to `Service Worker`).

src/components/grid-layer/grid-layer-component.jsx
aabdaab: thank you 💯

weberjavi:
@aabdaab does this PR has your approval? 😅

aabdaab:
Yeah, that's definitely better with the requests. I tried to do some digging in the performance tab in chrome:
image
as it has some bottlenecks (red bars) but I think it's a bigger topic and I don't want to stall that PR. I'd like to revisit that though, as I don't entirely understand what's so slow there 😅I just pinpointed that's something in `dojo` and has to do with scheduling, also not sure if we can manipulate that.

PR #103, opened 2019-09-09
Render tooltips on landscape mode for conservation areas

aabdaab:
Add tooltips for conservation areas on landscape mode.
1. Override esri's `popup` styles
2. When in landscape mode, fetch community and protected areas feature layers (once this is reviewed, we can remove them from the Web Scene)
3. Community tooltip has priority over protected tooltip, as their info overlap; 
4. Optimize querying the layers; if community area has point info, don't query protected layer

image


gretacv:
Hi @aabdaab,

I think it might make sense follow the same 'priority' logic than in the conservation widget, so, what do you think if 

> protected tooltip has priority over Community tooltip 
?

aabdaab:
but that would mean that when clicking on community area (yellow) you'd still get info about the protected area? (red area)

gretacv:
Maybe I misunderstood your comment, I thought that in cases yellow and red overlap you would give priority to community. 
Now that I am thinking of it... shouldn't the top layer have priority? I mean depending on the ordering in the legend?

aabdaab:
Also, the screenshot might be confusing because I couldn't activate the conservation layers as there are some changes being made on the data globe web scene, so I had to test it on featured globe web scene (and couldn't easily activate those layers).
And as you said, when community and protected are rendered, I'm giving priority to community - displaying yellow tooltip. But that's a good point you've made. If the legend is reordered we should take that into account...

weberjavi:
☝️ checking

weberjavi:
@aabdaab seems that this branch is introducing some deprecated code (is asking to install `react-helmet`, this dependency was already removed on the [caching PR](https://github.com/Vizzuality/half-earth-v3/pull/97)), please check that out and merge the latest develop version 🙏

aabdaab:
rebased with develop ☝️

fannycc:
> Maybe the tooltip should disappear once we get outside landscape mode? That'a a question for @fannycc probably.
I'm guessing the tooltip works on click and not on hover right? I can't appreciate that in the screenshot. Also the cursor should be a hand as Javi said to indicate affordance and that the area is clickable.
In my opinion the tooltip should disappear whenever user does another interaction, (click outside, zoom in, zoom out, drag map, go back to global view...) Otherwise we'd need to put an `x` icon to close it. 
Another thing that is confusing me a bit is why if the user is clicking in the same yellow area, I'm seeing 2 different names depending on where I'm clicking. (I think @gretacv and @aabdaab were discussing this before and I'm joining late this conversation).

aabdaab:
@fannycc The same yellow area probably consists of different protected places, but because we don't highlight their ranges to us it seems like it's the same area. We could think about when clicking and showing the tooltip, also highlighting the borders of that area.
As for the tooltip interaction - yes, it works on click. The tooltip disappears when clicking outside of the protected areas (when clicking in a different place but within the protected areas, the tooltip changes its place and content - it displays info about that clicked point).
I agree with other interactions (zooming in, out, leaving landscape mode); the tooltip should disappear, I'll add that.

PR #104, opened 2019-09-10
Feature/species widget

kfilip94:
This PR implements Species widget 
[PIVOTAL](https://www.pivotaltracker.com/story/show/168199334)

One question from me: do we want to keep tooltips after hovering on chart dot and also keep these dots clickable?

0qxx2-81bjv

**NOTE:** Please check if in Safari and Google Chrome  the problem with lack of background blur occurs (bug on MacOS related with child elements with `position: fixed`)


fannycc:
Looking good, I haven't designed the dots so close to the lines but I wasn't being realistic :) It's sad to see how many species are in the outskirts :( Nice job!
Regarding your question: I think it looks good as it is now. It makes sense to keep the hover so the user knows which species is before clicking on it...I'd love to hear other's thoughts. But i'm happy

weberjavi:
☝️ checking

kfilip94:
@fannycc I think this is the data issue, we don't have IUCN and also English names for every species, so sometimes I had to copy the scientific name into the place of the English name

PR #108, opened 2019-09-19
Add pledges layer to map iframe globe & fix unclickable entry box

aabdaab:
- Fix the first entry box (Biodiversity) being unclickable - the bug appeared as a side effect of css animations PR ;) 
- Fix rendering `map-iframe` the `/map` path - as I discovered it wasn't working since we introduced featured mode
- Add pledges layer to the `map-iframe` globe - add it as a `FeatureLayer` with `featureReduction` of `selection` type enabled. The layer is added on `handleMapLoad`, it's not living on the WebScene. FeatureReduction option allows for the pledges not overlapping each other. FeatureReduction redraws dynamically depending on the zoom level. The spinning effect wasn't added for now.
- To see pledges layer on the globe and test it locally change this line of code in `map-iframe.js`:
```
layer.visible = false; // change to true
```
and go to `/map` endpoint
When that layer will be activated by `postRobot` we're gonna have a legend as well. Not sure if that's desired, we can add the pledges layer to `LEGEND_FREE_LAYERS` array to avoid showing it there.

image

I'm also posting here the 3 methods we have for showing the pledges layer for reference. We went for the 2nd option without the spinning effect to avoid lights appearing/disappearing (blinking effect).
There are three ways of showing them:
1. show all the pledges at once. caveat is that there's a LOT in the US so they overlap each other and it looks weird. there's a border issue. the benefit of that is that while spinning the globe, the pledges are not being redrawn on the globe - they're fixed (meaning that they're not appearing/disappearing depending on the view port)
2. feature reduction. we don't allow pledges overlapping each other. depending on the zoom there are as many pledges of the same icon size as long as they don't overlap each other. no border issue. While spinning the globe though, the pledges are redrawn depending on the viewport; feature reduction depends on the viewport. the issue of appearing/disappearing emerges (maybe not the issue? see the videos in the dev channel)
3. clustering. different pledge sizes. sometimes border issue. spinning the globe makes the pledges being redrawn every 3 sec (I had to tweak the library) -  the issue of appearing/disappearing emerges

### DIFF
--- a/src/layouts/app-layout/app-layout-component.jsx
+++ b/src/layouts/app-layout/app-layout-component.jsx
@@ -2,13 +2,15 @@ import React, { PureComponent } from 'react';
 import Proptypes from 'prop-types';
 import FeaturedGlobe from 'pages/featured-globe';
 import DataGlobe from 'pages/data-globe';
-
+import MapIframe from 'pages/map-iframe';
 
 class App extends PureComponent {
   render() {
     const { route } = this.props;
     const { page } = route;
-    return page === 'featured-globe' ? <FeaturedGlobe /> : <DataGlobe />;
+    const embedded = page === 'map-iframe';
+    const whichGlobe = embedded ? <MapIframe /> : <DataGlobe />
+    return page === 'featured-globe' ? <FeaturedGlobe /> : whichGlobe;
   }
 }
 
--- a/src/assets/icons/pledge.svg
+++ b/src/assets/icons/pledge.svg
@@ -1,0 +1,14 @@
+<svg width="24" height="24" xmlns="http://www.w3.org/2000/svg">
+  <defs>
+    <filter x="-95%" y="-95%" width="290%" height="290%" filterUnits="objectBoundingBox" id="a">
+      <feGaussianBlur stdDeviation="3" in="SourceGraphic"/>
+    </filter>
+    <filter x="-100%" y="-100%" width="300%" height="300%" filterUnits="objectBoundingBox" id="b">
+      <feGaussianBlur stdDeviation="2" in="SourceGraphic"/>
+    </filter>
+  </defs>
+  <g transform="translate(7 7)" fill="none" fill-rule="evenodd">
+    <circle filter="url(#a)" cx="5" cy="5" r="5" fill="#18BAB4"/>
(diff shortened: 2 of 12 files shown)


weberjavi:
Checking ☝️

src/layouts/app-layout/app-layout-component.jsx
```diff
@@ -2,13 +2,15 @@ import React, { PureComponent } from 'react';
...
 import DataGlobe from 'pages/data-globe';
-
+import MapIframe from 'pages/map-iframe';
 
 class App extends PureComponent {
   render() {
     const { route } = this.props;
     const { page } = route;
-    return page === 'featured-globe' ? <FeaturedGlobe /> : <DataGlobe />;
+    const embedded = page === 'map-iframe';
```
weberjavi: Ok, this is working and I do not want to block this PR from being merged, but probably we will need to review the use cases for the embed map (as it seemed the other day that Chris is not using it any more, as he said to get rid of the `location` params under the share url).
Also the logic behind this page selection might not respond to the interactions currently available for the user on the platform (as we can also share the `iframe` on the `featuredGlobe` not only on the data one).
Maybe having a `embed` layout separated from the  from the `app-layout`?

aabdaab: so, maybe it is the naming, but I thought there are two different use cases here:
- the `iframe` we have in the Share modal that we can embed anywhere and do everything we can do on the website,
- the `map-iframe` that has a separate path `/map` that is for Chris to do some storytelling with it (+ the pledges) and use `postRobot` for that. For that map-iframe it's true, it's using only the expert mode web scene. We will probably have to revisit it anyway after we know more what kind of storytelling they'd like to do when it comes to the featured mode.

aabdaab:
Can I merge this? We'd probably have to review the map-iframe anyway as I indicated here:

```The iframe we have in the Share modal that we can embed anywhere and do everything we can do on the website (paths /featuredGlobe, /dataGlobe),
the map-iframe that has a separate path /map that is for Chris to do some storytelling with it (+ the pledges) and use postRobot for that. For that map-iframe it's true, it's using only the expert mode web scene. We will probably have to revisit it anyway after we know more what kind of storytelling they'd like to do when it comes to the featured mode.
```

weberjavi:
It is ok with me 👌

PR #109, opened 2019-09-19
Fix conservation pie logic

aabdaab:
This PR solves the bug of the grey pie slice sometimes being negative.

**Before**:
Screenshot 2019-09-19 at 10 07 33

**After**
image


weberjavi:
Checking ☝️

PR #110, opened 2019-09-19
Fix mosaic rule

weberjavi:
This PR fixes the human land pressures layer `mosaicRule` bug. The last selected set of land pressures rasters was persisted when changing experience (so if the last pressure selected on the `featured`globe was `urban` when changing to `expert` and selecting `rainfed` only `urban` raster will be displayed).

[PIVOTAL](https://www.pivotaltracker.com/story/show/168134824)


aabdaab:
checking ☝️

PR #111, opened 2019-09-19
Fix infinite loading spinner on activating conservation areas

aabdaab:
Sometimes we had an infinite loading spinner issue when activating protected areas layers as there was a bug in `layerToggled` function:
```
const layerToggled = map.layers.items.reduce((wantedLayer, currentLayer) => {
      if(wantedLayer) return wantedLayer; // without that line, even if we found the layer, we were still going through other layers and returned `undefined` instead, it was dependent on the layers order
      if(currentLayer.title === option.id) return currentLayer;
      if(currentLayer.layers) return currentLayer.layers.items.find(layer => layer.title === option.id);
      return wantedLayer;
    }, null)
```

### DIFF
--- a/src/components/protected-areas-layers/protected-areas-layers-component.jsx
+++ b/src/components/protected-areas-layers/protected-areas-layers-component.jsx
@@ -4,7 +4,7 @@ import PropTypes from 'prop-types';
 import MultipleActiveLayers from 'components/multiple-active-layers';
 
 import { usePaintLayer } from 'hooks/esri';
-import { handleLayerRendered } from 'utils/layer-manager-utils';
+import { handleLayerRendered, getToggledLayer } from 'utils/layer-manager-utils';
 import { WDPALayers, PROTECTED_AREAS_COLOR, COMMUNITY_AREAS_COLOR } from 'constants/protected-areas';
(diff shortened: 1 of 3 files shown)


weberjavi:
Checking ☝️

aabdaab:
done ☝️

src/components/protected-areas-layers/protected-areas-layers-component.jsx
```diff
@@ -25,12 +25,7 @@ const ProtectedAreasLayers = ({ handleGlobeUpdating, handleLayerToggle, activeLa
   const toggleLayer = (layersPassed, option) => {
```
weberjavi: I was talking about this outer `toggleLayer` function, not the inner `layerToggled`. As you can see there is more shared logic inside of it (updating the flag for displaying the loader, analytic events... )

aabdaab: right 🤦‍♀ on it

weberjavi:
Hey @aabdaab is this ready again?

aabdaab:
yep

weberjavi:
checking

weberjavi:
thanks!!

PR #113, opened 2019-09-20
Disable interactions on globe when featured place is selected

aabdaab:
Disable all interactions on the globe when the featured place is selected.
For more explanation look [here](https://www.pivotaltracker.com/story/show/168215369)

### DIFF
--- a/src/components/widgets/location-widget/location-widget.js
+++ b/src/components/widgets/location-widget/location-widget.js
@@ -13,7 +13,7 @@ import { clickFindMyPositionAnalyticsEvent } from 'actions/google-analytics-acti
 const actions = { ...urlActions, clickFindMyPositionAnalyticsEvent };
 
 const LocationWidget = props => {
-  const { view, changeGlobe, clickFindMyPositionAnalyticsEvent, isNotMapsList } = props;
+  const { view, changeGlobe, clickFindMyPositionAnalyticsEvent, hidden } = props;
   const [locationWidget, setLocationWidget] = useState(null);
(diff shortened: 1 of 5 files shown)


kfilip94:
:point_up: checking

kfilip94:
@aabdaab a few things to tweak:
After click on `Explore this area` interactions with the landscape sidebar are disabled.
Also, we can still interact with the globe by using the location button or zoom in/out widget, should we disable them as well?

There is a bug with an area image on the globe, the icon of the first area is it is still visible during the "tour" (check gif below). It is not a part of this PR, but this bug should be fixed somewhere:
wtwnf-m76fx

aabdaab:
I will check it after the core meeting

aabdaab:
```
Also, we can still interact with the globe by using the location button or zoom in/out widget, should we disable them as well?
```

Hmm, good point. I think we can remove them from the viewport the same as we do in the all featured maps list, wdyt?

aabdaab:
- Fixed the bug with landscape mode not being interactive
- Fixed the bug with an area image on the globe, the icon of the first area is it is still visible during the "tour"

Let me know what you think about not showing the widgets on the right side of the screen; location, zoom, zoom out, full screen?

kfilip94:
@aabdaab  thank you for changes, good job ;)
To me you can hide the navigation widgets, I think there is no point to show them in the "tour mode" if we're going to disable them anyway.

aabdaab:
done ☝️

kfilip94:
@aabdaab thanks!

PR #116, opened 2019-09-26
Add param which when true shows pledges on the map-iframe globe

aabdaab:
PT task: https://www.pivotaltracker.com/story/show/168779418
Show pledges layer on the map-iframe globe if
a) pledges slug is in `activeLayers`
b) `isPledgesActive` param is set to true
The new param has been added to facilitate user-friendly activation of the pledges layer.
Go to `/map?isPledgesActive=true` to test.
On default pledges layer is not visible so on `/map` we don't render pledges layer.


kfilip94:
checking

PR #120, opened 2019-10-01
Landscape mode mobile styles

aabdaab:
Landscape mode styles + behavior.
New flag added `isLandscapeSidebarCollapsed` for the landscape mode case on mobile; the behavior is as follows: landscape sidebar is always visible but it may be collapsed:
image
You can open all the options from the global view on the landscape mode on mobile. They will be placed above the collapsed landscape sidebar:
image
- When the landscape sidebar is open and we click on any option in the menu footer, we collapse the landscape sidebar.
- When we expand the collapsed landscape sidebar and any option from the menu footer is open, we close it and expand the landscape sidebar.

All the mobile styles are still work in progress so it might be that some styles are yet to be tweaked; nevertheless, let me know in the comments.


fannycc:
I don't understand the last image... Whenever user clicks on the bottom bar the options should be displayed as always, as it's where we have the focus: see example below: 
HE Mobile - Landscape closed Copy

fannycc:
regarding styles I don't have any comment, all looks ok

fannycc:
yes..because they act like switches that appear on top (not above) of anything else in the screen. I see the confusion here with the word on top :( so yes, they'd replace it. Let me know if you need anything, we can chat or if It's a hard change so we maybe can leave it like this for now.

aabdaab:
yeah, it's a tricky word 😅it's a quick enough change, I'll replace the landscape sidebar with anything that was clicked on the menu footer; unclicking would bring the landscape sidebar again to the view.

fannycc:
thanks! that behaviour will work :)

kfilip94:
:eyes: checking

kfilip94:
- Hide / move  tutorial in landscape mode:
image

- Disable inside scroll when sidebar is collapsed:
34kiu-aif8y

- Increase the left and the right padding of the sidebar content to 32px.
image

- Top border-radius works weird when scrolling, I think it should work in the opposite way:
oqdmt-28l8w

- I think it would be better to have a clickable title, not only this small arrow, click on the title should also show/hide the sidebar: 
lb9o1-b4d1g

fannycc:
can the headers of the sidebars be fixed so users can go back or collapse in any moment?

src/components/mobile-only/menu-footer/menu-footer.js
```diff
@@ -45,7 +48,14 @@ const MenuFooterContainer = props => {
 
   const handler = (option) => {
     if (activeOption === option) resetActiveOption();
-    else setActiveOption(option);
+    else {
+      if (isLandscapeMode && !isLandscapeSidebarCollapsed) {
```
kfilip94: Here we can do it a bit simpler ;)

```suggestion
      isLandscapeMode && !isLandscapeSidebarCollapsed && collapseLandscapeSidebar();
      setActiveOption(option);
```

aabdaab: this is removed in another PR

aabdaab:
Ready!

aabdaab:
closing this, there's an identical branch here checking out from a correct branch: [here](https://github.com/Vizzuality/half-earth-v3/pull/127)

PR #126, opened 2019-10-02
Feature/scene initial load

weberjavi:
Will explain everything that's happening here in a comment once this is ready for review :)


weberjavi:
Ok, so, this is done. I'll need to fix the conflicts with the recently merged code (the bug fixes PR) but feel free to start testing this beast out and reviewing the code. Probably some further polishing is needed, but I've tried to be as clean as my brain was able to be, I've touched way too much code... 82 files :(

The main thing I was trying to accomplish here was to reduce load times for the scene, I think I've reached the goal. I changed  the underlying `component`, now instead of the `webScene` from `react-arcgis` I've created a custom `scene` through `esri-loader`. So instead of fetching all at once we just fetch the `layers` as needed.

I've added some `constants` files to store layers slugs, urls, configurations and what not. We need to define the layer `type` on the layer config to let the app know which esri module should handle the layer creation (if no `type` is provided a `WebTileLayer` constructor is used. This is the `type` of the layers we are already fetching from google MoL server.)
We are now using layers stored on ESRI servers for almost all the biodiversity layers (but the fishes, fine scale and some other)

Also added some utility functions to the `layer-manager-utils` to handle all the new flow (checking if layer is on the map, layer creation and all that stuff).

I've debugged and refactored a bit along the way. This are the main pivotal issues tackled on this pr:
-  [Scene load improvements](https://www.pivotaltracker.com/story/show/168213854)
-  [Move species widget](https://www.pivotaltracker.com/story/show/168536112)
-  [Human pressures colour palette update](https://www.pivotaltracker.com/story/show/168895709)
-  [Update selected taxa icon colour](https://www.pivotaltracker.com/story/show/168919398)

I think also this three pieces of feedback are working:
-  _Are the "center" and "zoom" options needed here? (see attached screenshot: H_E_map_share.png)_
-  _The link from the Share text doesn't do anything (actually, there is an error in the console, so might need to fix this). Sample link._
-  _The HELP icon on the legend doesn't seem to do anything, apart from pulsing_

Gonna update the demo site for you to be able to test this online
Again sorry for opening this huge of a PR 🙇

kfilip94:
:mag:  checking

aabdaab:
hey, I've noticed you enabled the `all-taxa-rarity` layer on the first load: we're still missing the number next to the Biodiversity entry box indicating that there's a layer activated + the opacity is set to 1 instead of 0.6:
image

kfilip94:
`About the Half-Earth`doesn't appear on data globe:
image

src/components/labels-layer/labels-layer.js
```diff
@@ -1,3 +1,78 @@
...
+  });
+}
+
+const LabelsLayer = props => {
+  const { map } = props;
+
+  useEffect(() => {
+    loadModules(["esri/layers/support/LabelClass"])
+    .then(([labelClassConstructor]) => {
+      var style_1_ter = labelClassFactory(labelClassConstructor, 'style_1_ter');
```
kfilip94: It's more like a suggestion, but we use here an array of slugs, like:
```suggestion
      const slugs = [ 'style_1_ter', 'style_2_ter', 'style_3_ter', ... ];
      const labelClasses = slugs.map(slug => {
        return labelClassFactory(labelClassConstructor, slug);
      });
      
      // so at the end:
      
      (...)
      .then(layer => {
            layer.opacity = 1;
            layer.visible = true;
            layer.labelsVisible = true;
            layer.labelingInfo = labelClasses
      })
      (...)
      
```

weberjavi: Indeed 👏 
Nice one

weberjavi:
Hey @aabdaab , @kfilip94 addressed your comments and suggestions and also added[ this task ](https://www.pivotaltracker.com/story/show/168774291)on this same branch (add the RAISIG vector tile layer to the map when displaying the Community protected areas).
Ready again 👊

kfilip94:
@weberjavi It seems to be working nice! I think we should merge it.
Thanks for your effort, you did an awesome job here, it's noticeably faster! :trophy: :rocket: :man_mechanic:

PR #134, opened 2019-10-05
Fix/add loading loaded flags to sidebar components

kfilip94:
This PR clears content inside Species widget and Conservation Efforts widget when new data is being fetched.
5oxal-xsz3b


aabdaab:
Let's leave it for Monday. I saw some weird edge cases like an empty conservation widget without any explanation. We should thoroughly test it.

kfilip94:
ready again :point_up:

aabdaab:
checking ☝️

aabdaab:
It's not really related to this PR but since I was checking the widgets on the landscape mode on mobile, could you fix the blur on the species widget?:
image
I think adding `<DummyBlurWorkaround />` as an immediate child of a container should fix this.

kfilip94:
fixed, check this again please because this bug appears only on MacOS

PR #138, opened 2019-10-07
Fix broken layers counter in protected areas entry box

kfilip94:
This PR fixes the incorrect number of selected layers in `Existing protection`: when the `Community-based` checkbox was selected, the number of currently active layers was greater by two, not by one. The reason is that selecting `Community-based` layer now selects two layers: `protected_areas_vector_tile_layer` and `RAISIG_areas_vector_tile_layer`

### DIFF
--- a/src/components/entry-boxes/entry-boxes-selectors.js
+++ b/src/components/entry-boxes/entry-boxes-selectors.js
@@ -9,7 +9,10 @@ const getCountedActiveLayers = createSelector(
   (activeLayers, rasters) => {
  
     const biodiversityLayers = activeLayers ? activeLayers.filter(({ category }) => category === 'Biodiversity').length : 0;
-    const protectionLayers = activeLayers ? activeLayers.filter(({ category }) => category === 'Existing protection').length : 0;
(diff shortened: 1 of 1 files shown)


src/components/entry-boxes/entry-boxes-selectors.js
```diff
@@ -9,7 +9,10 @@ const getCountedActiveLayers = createSelector(
   (activeLayers, rasters) => {
  
     const biodiversityLayers = activeLayers ? activeLayers.filter(({ category }) => category === 'Biodiversity').length : 0;
-    const protectionLayers = activeLayers ? activeLayers.filter(({ category }) => category === 'Existing protection').length : 0;
+    const protectionLayers = activeLayers ? activeLayers.filter(({ category, title }) => {
+      // we have to filter 'RAISIG' layer because activating 'Community-based' checbox selects two layers on the globe: "protected_areas_vector_tile_layer" and "RAISIG_areas_vector_tile_layer"
+      return category === 'Existing protection' && title !== 'RAISIG_areas_vector_tile_layer'
```
aabdaab: you might want to use `RAISIG_AREAS_VECTOR_TILE_LAYER` const from layer-slugs

PR #141, opened 2019-10-08
Preserve zoom level on refreshing the page

aabdaab:
This PR tackles the problem of landscape mode not being preserved when sharing the link or refreshing the page. The issue laid in FeaturedPlaceViewManager which always would set zoom to 1.
Now, it sets the zoom which is preserved in the url.

[PT](https://www.pivotaltracker.com/story/show/168480884)

### DIFF
--- a/src/components/featured-place-card/featured-place-card.js
+++ b/src/components/featured-place-card/featured-place-card.js
@@ -56,7 +56,10 @@ const FeaturedPlaceCardContainer = props => {
     }
   }, [featuredPlacesLayer, selectedFeaturedMap, selectedTaxa])
 
-  const handleAllMapsClick = () => changeUI({ selectedFeaturedPlace: null });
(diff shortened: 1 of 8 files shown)


kfilip94:
When trying zoom out from the landscape mode, the camera goes back to the previous zoom level :(

7ymu8-bwh5v

aabdaab:
Ready again ☝️ 
I moved the `view.goTo` to a click handler - removing it from useEffect fixes bugs related to different desired behaviors while having the same redux state, i.e.
```
if (!isLandscapeMode) { view.goTo(zoom: 1 || zoom) }
// we wanted to execute that only when leaving the featured place card (clicking back button)
// but it was executed also when having a zoom < 8.
```
We should avoid depending on a state when it might be ambiguous and we expect to have different behaviors while having the same state.

PR #146, opened 2019-10-10
Add email icon to share modal

aabdaab:
This PR adds an email option to the share modal.
[PT](https://www.pivotaltracker.com/story/show/169062509)

image

**To note**:
There's an issue with the text limit that is passed to `mailto:` body content. See for example, when you go to the featured place card and try to share from there, you'll see that the body in the email is stripped (in my case it's not taking `ui=...` into account. Can you check if you run into a similar problem? I have seen a thread about this in [Stackoverflow](https://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers) although it mentions a lot bigger character limits (~2000 chars) whereas ours is less than 150 chars... Either way, I think we should add another task for researching how can we hash the URLs to make them shorter.


weberjavi:
☝️ checking

src/components/share-modal/share-modal-component.jsx
```diff
@@ -59,15 +59,13 @@ const ShareModal = ({ handleClose, isOpen, shareSocialMedia }) => {
       </div>
       <div className={styles.socialMediaContainer}>
         {shareSocialMedia.map(socialMedia => (
-          <Button
+          <button
             onClick={() => window.open(`${socialMedia.link}${currentLocation}`)}
```
weberjavi: We can avoid the lost of the params after the ampersand by encoding the URL before we send it over the to the given social media:
`onClick={() => window.open(`${socialMedia.link}${encodeURIComponent(currentLocation)}`)}`

aabdaab:
☝️ fixed!

PR #150, opened 2020-01-29
Feature/user data modal

weberjavi:
This PR implements a modal to gather Users data in order to drive the next user research phase.
An [airtable](https://airtable.com/) service has been created to store the data for further use of the research team.

Two new ENV variables need to be added to the project:
`REACT_APP_AIRTABLE_API_KEY` and `REACT_APP_AIRTABLE_USER_RESEARCH_BASE_ID`

[DESIGN](https://projects.invisionapp.com/d/main#/console/17484384/401570096/preview)

The user only should be able to see the modal the first time she visits the app, so a local storage `key` is set in order to check that.


kfilip94:
checking :eye: :pig_nose: :eye:

kfilip94:
Hey @weberjavi, I found a few issues here, you can find them below. I'm checking the code now ;)
- After click on the `x` button, app breaks:
k88y7-7rtc6

- Mobile version needs to be polished:
image

- Contact input from the second step behaves a bit weird, it shares the text with the profession input:
32h2j-38e58

- Shouldn't be `Write ...` instead of `Writte ...`?
image

- I'm not sure if we should allow for sending `null` to the server that easily:
79825-g8hop (1)

src/components/user-data-modal/user-data-modal-component.jsx
```diff
@@ -0,0 +1,98 @@
...
+      <p className={styles.question}>Please tell us what you use the map for:</p>
+      <input type='text' className={styles.textInput} placeholder={'What do you use the map for?'} onChange={(event) => userDataUpdate('map_usage', event.target.value)}/>
+      <button className={styles.button} onClick={() => setActiveStep('step2')}>SEND</button>
+      </section>
+    )
+  }
+
+  return (
+    <Modal
+    onRequestClose={userData => handleModalClose(userData)}
```
kfilip94: :pray:  please add white spaces here :D

weberjavi:
Thanks @kfilip94 will address this bits tomorrow 🙇

weberjavi:
Hey @kfilip94 ready again. Take into account that there's an issue with `CSS` assets ordering on `development` so the `he__modal-styles` from the `he-components` repo overwrite the custom styles from the project. This problem gets solved after building the app for production.

Check the gif with the proper styles precedence.

Kapture 2020-02-05 at 13 00 12

Thanks for the review!!

kfilip94:
Hey @weberjavi thanks for changes :+1: 
About `I'm not sure if we should allow for sending null to the server that easily:` I wasn't very descriptive, sorry.  I meant that we probably shouldn't send a request at all if all the fields are empty:
image

It's more a suggestion because leaving that can add a lot of noise and empty rows to the data that our researchers need to analyze at the end. I'm also not sure how Airtable deals with empty rows, so maybe it's not a problem at all. What do you think about that?

weberjavi:
Yeah, maybe you're right. Gonna change the way the request are handled and perform two requests to the table, one with the `profession` and `map usage` info (on the first step of the form) and another one on the second step with the `email` associated to the previously created user (that way will remove the request from the closing button).

Anyway airtable allows you to filter rows empty values, which is useful for this kind of forms without required fields.

Will let you know when this is ready ✊

weberjavi:
Ready again @kfilip94 ☝️

kfilip94:
Hey @weberjavi, I assume it happens only because I don't have any key for Aritable in my .env file, right?
dcgof-j26en

weberjavi:
☝️ ready

kfilip94:
Hey, @weberjavi thank you so much for your changes, everything works great! Let's merge it! :tada: :+1:

PR #154, opened 2020-03-02
Feature/add zeit integration

weberjavi:
This PR adds the `now.json` file needed to leverage `Zeit/Now` deploy functionalities on the Half Earth Project.
Once this PR is merged pushed branches to the repo will have an automatic deploy.


weberjavi:
We would need to whitelist also the Zeit url in order to be able to display the non public layers managed by MoL on this demo branches 🤔 (not sure if that would be possible, as the urls are dynamic).
But this will allow us to test other UI updates and the public layers managed by us (protected areas and human encroachment).
Will research the auth flow to check if it is possible to get the layers on the map after we enter our credentials.
I'm happy merging this though as it will improve some of our use cases.

PR #159, opened 2020-03-04
Feature/add area size on top of geodescriber description

kfilip94:
This PR adds the `Area` line on top of the geodescriber text:
[PIVOTAL](https://www.pivotaltracker.com/story/show/171629244) | [PREVIEW](https://half-earth-v3-c657cxzv4.now.sh/featuredGlobe?globe=eyJ6b29tIjozLjE3MTAyMjg0MTU4MTE0OTQsImNlbnRlciI6WzE2Ljk1MTU1MzYwMDAwMDA2OCwwLjExNjk1OTAwMTQzMTM1NTAxXX0%3D)

image

Also, I updated the selected layers counter styles, requested by fanny:
[PIVOTAL](https://www.pivotaltracker.com/story/show/169534134)

image

### DIFF
--- a/src/components/category-box/category-box-styles.module.scss
+++ b/src/components/category-box/category-box-styles.module.scss
@@ -48,11 +48,25 @@
   margin: 0;
   margin-bottom: 8px;
   user-select: none;
+  display: flex;
+  align-items: baseline;
+}
 
-  .counter {
-    color: $robins-egg-blue;
-    margin-left: 6px;
-  };
+.counter {
+  align-items: center;
+  background-color: $robins-egg-blue;
+  border-radius: 8px;
+  color: $firefly;
+  display: flex;
+  font-weight: $font-weight-bold;
+  height: 16px;
+  justify-content: center;
+  margin-left: 6px;
+  width: 16px;
+}
+
+.counterText {
+  height: 10px;
 }
 
 .categoryContainer {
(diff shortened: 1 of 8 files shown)


fannycc:
Quick comment, is there 24px of margin at the bottom of the geodescriber? just double-checking since it looks like 20px to me

fannycc:
> @fannycc You're right, nice catch! it's 20px indeed, I'll change it to 24px 👍

Perfect! check the other margins of the card just in case plz. Thanks!!

kfilip94:
I tweaked the left padding of the back button, now it's aligned with the title, and I also increase the bottom padding of the title to 20px (14px before). @fannycc have you noticed something else?
image

src/components/category-box/category-box-styles.module.scss
```diff
@@ -48,11 +48,25 @@
   margin: 0;
   margin-bottom: 8px;
   user-select: none;
+  display: flex;
+  align-items: baseline;
+}
+
+.counter {
+  align-items: center;
+  background-color: $robins-egg-blue;
```
weberjavi: I see that this is the colour previously used but I would suggest to use `$brand-color-main` (that is the `$java` colour instead of `$robins-egg-blue`).
Also the project colour variable `$dark-text` instead of `$firefly`.
That way if we apply platform wide colour changes on the projects variables those would be applied.
```scss
background-color: $brand-color-main;
color: $dark-text;
```

kfilip94: @weberjavi thanks for catching that, ready again

PR #161, opened 2020-03-09
Feature/update conservation widget with 55km data

kfilip94:
This PR updates the conservation efforts widget with new 55km data. The conservation efforts data requires double-checking, but this feature can be merged independently.
[PIVOTAL](https://www.pivotaltracker.com/story/show/171566937)

### DIFF
--- a/src/components/landscape-sidebar/conservation-efforts-widget/conservation-efforts-widget.js
+++ b/src/components/landscape-sidebar/conservation-efforts-widget/conservation-efforts-widget.js
@@ -72,7 +72,7 @@ const ConservationEffortsWidget = (props) => {
   useEffect(() => {
     if (terrestrialCellData && queryParams) {
       setConservationEfforts({ data: null, loading: true });
-      queryParams.where = `ID IN (${terrestrialCellData.map(i => i.CELL_ID).join(', ')})`;
+      queryParams.where = `ID IN (${terrestrialCellData.map(i => i.ID).join(', ')})`;
       conservationPropsLayer.queryFeatures(queryParams).then(function(results){
         const { features } = results;
         setConservationEfforts({ data: features.map(c => c.attributes), loading: false });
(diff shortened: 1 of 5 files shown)


src/selectors/grid-cell-selectors.js
```diff
@@ -7,7 +7,7 @@ export const getTerrestrialCellData = createSelector(
   [selectCellData],
   cellData => {
     if (!cellData) return null;
-    return cellData.filter(c => c.ISMARINE === 0);
+    return cellData;
```
weberjavi: Right now we are using two grids to create the aggregation of grid cells. The new 55km grid only stores terrestrial data. We are using the old grid to get only marine grid cells IDs (this situation will persists until MoL provides a dedicated marine grid). We do not have marine protection information right now (please @gretacv correct me if I'm wrong), so right now you need to filter terrestrial data before consuming it on the widget.

```js 
return cellData.filter(c => c.ID);
```
This snippet works because the old grid features do not have an `ID` attribute but a `CELL_ID` so getting only the ones with `ID` creates a terrestrial only subset.

src/components/landscape-sidebar/conservation-efforts-widget/conservation-efforts-widget.js
```diff
@@ -72,7 +72,7 @@ const ConservationEffortsWidget = (props) => {
   useEffect(() => {
     if (terrestrialCellData && queryParams) {
       setConservationEfforts({ data: null, loading: true });
-      queryParams.where = `CELL_ID IN (${terrestrialCellData.map(i => i.CELL_ID).join(', ')})`;
+      queryParams.where = `ID IN (${terrestrialCellData.map(i => i.ID).join(', ')})`;
```
weberjavi: If you do not filter out the marine grid cells on the selector when querying the layer on coastal zones (with terrestrial and marine cells) will not return any data (because the resulting array of the mapping will have some `undefined` elements.

src/selectors/grid-cell-selectors.js
kfilip94: Thanks! :+1:  I wasn't fully aware of that

kfilip94:
Hey @weberjavi, I rebased this branch with the recent human-encroachment changes on `epic/update-data-layers`, it's ready to check and merge

kfilip94:
Conservation widget data need to be doublechecked, it has been addressed in [this Pivotal task](https://www.pivotaltracker.com/story/show/171836483)

PR #163, opened 2020-03-11
Feature/blend marine and terrestrial layers

kfilip94:
This PR implements fetching the marine layers from the old layer `ConsProp`. It follows this thread started on Slack. 
So what I did here:
1. Fetching old `ConsProp` layer,
2. Querying the features from this layer,
3. Turn `marine` and `terrestrial` query into promises, and use `Promise.all()` to resolve them,
4. Save all layers to one `conservationEffortsData` state,
5. Calculate conservation efforts data depending on the structure of the cell (old or new)

The problem is that even when we fetch the marine data,  `RAISG_prop` `WDPA_prop` and `all_prop` are equal to zero, not sure why.


weberjavi:
@gretacv can you confirm this? Seems that we do not have summery data for marine protected areas.

PR #164, opened 2020-03-11
Feature/human encroachment data update

weberjavi:
#### Releated stories tickets:
[PIVOTAL](https://www.pivotaltracker.com/story/show/171566335)

Be aware that the Human pressure layers used here are not yet definitive and do not have the whole zoom range (the only display in zooms 1 to 6 so they are not visible on landscape mode). Also they do not have the correct color ramp defined on the [UI kit](https://projects.invisionapp.com/d/main#/console/17484384/365514646/preview)

The original intent of this PR was to update the terrestrial Human encroachment layers to use the new tiled layers stored on EOWBF ArcGIS online account; also the data displayed on the Human pressures landscape widget has been update tho show the new summarie data calculated with the new 55km terrestrial grid. 
To do this I've removed all functionalities related with `image layers` and `raster` rendering. This had implications all through the app since there where _ad hoc_ functionality related to this type of layer (`humanPressuresPreloadFixes`, `mosaicRuleFix`es and raster functions) that where triggered in all `pages`.

There are some concrete use cases here regarding layers and the legend that are forcing us to do several changes on functionality:

- single layer that gets activated and displays on the legend as it is. It is part of the `activeLayers` on the location and updating it's attributes through the legend only affects its behaviour.
- Single layer pertaining to a grouped legend item set of layers. Gets activated from a dedicated toggle and is part of the location `activeLayers` but has a grouped behaviour on the legend. Needs to have a common behaviour through all the layers on the map (opacity, order...). Any action on the legend should update all of them, and when a new layer from the group is added to the map needs to inherit the groups attributes (regarding ordering and opacity). New Human pressures lasyer are an example of this behaviour.
- Composed layer (by two or more layers), they are diferent items on the database but need to behave as one. They are treated separately on the `activeLayers` but get activated and deactivated on the same button and their properties (order, opacity...) must behave consistently. Community areas layers are an example of this behaviour.

The pattern followed to update the data behind the `human-pressure-widget` has been the same as the one in place for the `prtected-areas-widget` and the `species-widget`. Creating a dedicated `store` that gets updated whenever the agreggated grid cell data is updated (when that happens a query to the `GRID_CELLS_LAND_HUMAN_PRESSURES_PERCENTAGE` layer is performed to get the summary of the given grid cells).

Some other small changes have been made to the `layer-manager-utils`.


kfilip94:
checking :eyes:

src/components/arcgis-layer-manager/arcgis-layer-manager.js
```diff
@@ -9,28 +9,19 @@ const ArcgisLayerManager = ({ map, activeLayers, customFunctions }) => {
...
   
   useEffect(() => {
+    items.forEach(sceneLayer => {
+      const isVisible = activeLayers.some(activeLayer => activeLayer.title === sceneLayer.title)
+      if (isVisible) {
+        sceneLayer.visible = true;
+        setOpacity(sceneLayer, activeLayers);
+      } else {
+        sceneLayer.visible = false;
+      }
```
kfilip94: A small suggestion here, shorter syntax:
```suggestion
      sceneLayer.visible = isVisible;
      if (isVisible) { setOpacity(sceneLayer, activeLayers); }
```

src/components/human-impact-layers/human-impact-layers-component.jsx
```diff
@@ -1,39 +1,31 @@
...
 import PropTypes from 'prop-types';
 // Constants
-import { VIEW_MODE } from  'constants/google-analytics-constants';
-import { layersConfig } from 'constants/mol-layers-configs';
-import { LAND_HUMAN_PRESSURES_IMAGE_LAYER } from 'constants/layers-slugs';
+import { layersConfig, LAYERS_CATEGORIES } from 'constants/mol-layers-configs';
 import { humanPressuresLandUse } from 'constants/human-pressures';
 // Utils
-import { handleLayerRendered, layerManagerVisibility, handleLayerCreation } from 'utils/layer-manager-utils';
-import { humanPressuresPreloadFixes, dispatchLandPressuresLayersAnalyticsEvents } from 'utils/raster-layers-utils';
```
kfilip94: We don't use functions like `handleLayerRendered`, `toggleWDPALayer` or  `getToggledLayer` from  'utils/layer-manager-utils.js' anywhere, do you think we can remove them safely?

kfilip94:
Select some human pressure layers, refresh the page, select another h-p layer - they will get a default opacity instead of the opacity previously set for other layers:
7fyxs-3ulhm

src/components/legend/legend-selectors.js
```diff
@@ -68,8 +67,40 @@ const parseLegend = (config) => {
...
 
 const joinAgricultureTitles = (titles) => {
-  const trimmedTitles = titles.map(title => title.split(" ")[0]);
-  return `${trimmedTitles.join(' and ')} agriculture`; 
+  const nonAgricultureTitles = titles.filter(title => !title.toLowerCase().endsWith('agriculture'))
+  const agricultureTitles = titles.filter(title => title.toLowerCase().endsWith('agriculture'));
+  const trimmedTitles = agricultureTitles.map(title => title.split(" ")[0]);
+  const formattedAgricultureTitles = `${trimmedTitles.join(' and ')} agriculture`
+
+  return `${[...nonAgricultureTitles, formattedAgricultureTitles].join(' and ')}`;
```
kfilip94: Since we have four layers, we should tweak this method a bit:
```suggestion
  const nonAgricultureTitles = titles.filter(title => !title.toLowerCase().endsWith('agriculture'))
  const agricultureTitles = titles.filter(title => title.toLowerCase().endsWith('agriculture'));
  const trimmedTitles = agricultureTitles.map(title => title.split(" ")[0]);
  const formattedAgricultureTitles = `${trimmedTitles.join(' and ')} agriculture`;
  const formattedNonAgricultureTitles = `${nonAgricultureTitles.join(', ')}`;
  
  if(!formattedNonAgricultureTitles) return formattedAgricultureTitles;
  return `${[formattedNonAgricultureTitles, formattedAgricultureTitles].join(', ')}`;
```

What I did here is the separation with ` AND ` only the agriculture pressures, and anything else with a comma. The main confusion is that with separating everything with ` AND  ` could give us a false impression that each pressure belongs to the agriculture group:
**BEFORE:**
image

**AFTER:**
image

Makes sense?

gretacv: just curiosity, what happens when urban, rangeland and irrigated agriculture are on?

kfilip94: All would be separated with a comma:
image

gretacv: Is it too much to always have an "and" before the last element?

kfilip94:
In case when urban, rainfed and irrigated layers are selected, the legend clips the label:
image

src/components/legend/legend-selectors.js
kfilip94: I don't think so, but imagine this case:
we've selected three layers:
`RANGELAND, RAINFED AND IRRIGATED AGRICULTURE`
which is fine because layers names are "Rainfed agriculture", "Irrigated agriculture" and "Rangeland"
but if we unclick the "Rainfed agriculture" layer, we will get:
`RANGELAND AND IRRIGATED AGRICULTURE`
what seems to be confusing because we don't have ""Rangeland agriculture" layer, only "Rangeland"
Am I overthinking this? :D

gretacv: I think it is not a problem to have `rangeland and irrigated agriculture`

kfilip94:
Hey @weberjavi, well done, you did a really good job with replacing raster layers!
I left a few comments above, please check them whenever you can ;)

weberjavi:
Hey @kfilip94  thank you so much for the review!!!
I've committed your suggestions (👨‍🎨 ) and addressed all of your comments. The main thing I've done, beyond styles changes, has been `await`ing for the `handleLayerCreation` function to add the layer to the `scene` to proceed to update the `activeLayers`, this was the race condition that was avoiding the propper opacity to be set after reload.

Thanks, again 👍

PR #167, opened 2020-03-17
Feature/update species widget

weberjavi:
[PIVOTAL](https://www.pivotaltracker.com/story/show/171566930)

This PR adapts the landscape Species widget to the new aggregated grid cells and uses the recomputed species data (driven by the new 55km grid).

image

### DIFF
--- a/src/components/landscape-sidebar/species-widget/species-widget-selectors.js
+++ b/src/components/landscape-sidebar/species-widget/species-widget-selectors.js
@@ -1,5 +1,5 @@
 import { createSelector, createStructuredSelector } from 'reselect';
-import { getTerrestrialCellData } from 'selectors/grid-cell-selectors';
+import { getTerrestrialCellData, getMarineCellData } from 'selectors/grid-cell-selectors';
 import { format, precisionPrefix, formatPrefix } from 'd3-format';
 import { uniqBy } from "lodash";
 import IUCNList from 'constants/iucn-list';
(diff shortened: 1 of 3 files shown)


weberjavi:
@kfilip94 ☝️ this is ready to review

weberjavi:
conflicts resolved ☝️

weberjavi:
☝️ ready again

kfilip94:
Hey @weberjavi, I've noticed that each time that we pass the marine cells into the data query, the `speciesLayer.queryFeatures(query)` request will responses with an error:
image
This means that even if we have data available for terrestrial cells, we'll not display it because marine cells ID passed to `where` clause breaks the query. Should we exclude marine cells from the request?

weberjavi:
🤔 that's not happening in my local. I have `force pushed` previously, are you sure you have the last version of the branch?
image

kfilip94:
I think everything is up to date on my machine. Try to perform it on some island area, Tenerife for example - when the marine cells id are passed to the query, it throws an error (I print this error by adding the `.catch(err => console.error(err))` clause,  I forgot to mention that, sorry). But once we zoom in on the terrestrial cell, it works:
dxam8-8jcpt

However, I also checked the area that  you're checking, and everything seems to be all right :man_shrugging:
image

weberjavi:
@kfilip94 finally I removed the marine cells from the query. Since we are not getting marine species data that would make no difference on the data displayed on the widget and would avoid the error.
It seems that the error was being triggered due to the formatting of the marine cells ids, the id is generated with positive and negative numbers depending on the quadrant the cell is in, we will need to take this into account whenever we get the final marine grid.

PR #168, opened 2020-03-18
Feature/add country layer to the scene

weberjavi:
[PIVOTAL](https://www.pivotaltracker.com/story/show/171810267)
This PR adds the country labels feature layer to the `data globe`.
The initial approach to handle the countries data will be similar to the one we are using with the grid cells. The [layer with the labels](https://eowilson.maps.arcgis.com/home/item.html?id=46e7cb3493024df0bd978b15106dfaf9) holds a reference to the name (`NAME_0`) and the ISO code of the country (`GID_0`). With that reference we would be able to query the [countries data layer ](https://eowilson.maps.arcgis.com/home/item.html?id=b703def0c9724b689100b0b62ce6cef0)(that is quite heavy) without the need of adding it to the map.


weberjavi:
After the designs I'm assuming it only applies to the data globe.

weberjavi:
merginggg >>>>>

PR #169, opened 2020-03-18
Add events to the labels layer

weberjavi:
[PIVOTAL](https://www.pivotaltracker.com/story/show/171812076)

This PR adds the needed event listeners to the Country labels layer.
Right now the `countryISO` param on the URL `location` params is updated whenever a user clicks on a country label.

Kapture 2020-03-18 at 16 59 37


kfilip94:
I forgot to mention that: please change `country-labels-layer-component` file type from `.js` to `.jsx`

PR #171, opened 2020-03-23
Feature/local scene rendering

kfilip94:
This is the first iteration of the local scene rendering feature. I added `DoubleScene` component which is a modified version of the old `Scene` component - it renders two scenes, local and global, with two separate scene views, but with one shared map. Adding react portal feature enables using the same components in two different scenes without changing the logic that we already have.
[PIVOTAL](https://www.pivotaltracker.com/story/show/171811037)

To switch scene mode, you can click on `TOGGLE SCENE MODE` button, it's a temporary component for development purposes:
ebebx-coamw

We still need to work on how this component looks, fetching a country extent. 

KNOWN ISSUES:
- Search component doesn't work properly on the local scene,
- zoom and fullscreen buttons disappear on the local scene, we need to apply them to the local scene view as well


weberjavi:
checking ☝️

PR #172, opened 2020-03-25
Feature/country entry card

weberjavi:
This PR adds the `country entry card` component to the sidebar. It also handles the removal of the country borders when the country is deselected.
I've left out of this task the mobile styles since it seems that this card it is supposed to live inside the `find places` widget. Some further clarifications of the flow from design will be needed.

[PIVOTAL](https://www.pivotaltracker.com/story/show/171810743)


weberjavi:
Gonna remove this PR in favour of https://github.com/Vizzuality/half-earth-v3/pull/173

PR #179, opened 2020-04-07
Feature/adjust local scene styles

kfilip94:
This PR adds the countries mask layer to the local scene.
[PIVOTAL](https://www.pivotaltracker.com/story/show/172018109) | [PREVIEW](https://half-earth-v3-git-feature-adjust-local-scene-styles.vizzuality1.now.sh)
om536-hsi2g

It also:
- increases the size of the red country marker,
- fixes missing marker after the first load of the page,

- sets the maximum tilt to 60,
- adds the left padding to the local scene,

Things that still need to be fixed:
- Sometimes when we click on the new country label and then quickly on "Explore this country", the country mask on the local scene doesn't appear - something is wrong with the layer visibility, haven't discovered what exactly,
- defining camera constraints for the local scene seems to be more complicated - the altitude constraints don't work on local scenes, also can't constrain the panning. I can only disable it completely, what as I understand it's not desired behavior either - need to dig deeper, maybe there is another solution for that
- Smarter elevation - for smaller countries the elevation level goes crazy, will check it in a different PR

### DIFF
--- a/src/components/country-border-layer/country-border-layer.js
+++ b/src/components/country-border-layer/country-border-layer.js
@@ -5,8 +5,6 @@ import { LAYERS_URLS } from 'constants/layers-urls';
 import { gridCellDefaultStyles } from 'constants/landscape-view-constants';
 import { connect } from 'react-redux';
 import actions from 'redux_modules/country-extent';
-import { simplePictureMarker } from 'utils/graphic-layer-utils';
-import mapPinIcon from 'icons/map_pin.svg'
 
 import {
   createGraphic,
@@ -15,23 +13,17 @@ import {
 } from 'utils/graphic-layer-utils';
 
 const CountryBorderLayer = props => {
-  const { view, spatialReference, countryISO, setCountryExtentLoading, setCountryExtentReady, setCountryExtentError, isCountryMode } = props;
+  const { view, spatialReference, countryISO, setCountryExtentLoading, setCountryExtentReady, setCountryExtentError } = props;
 
   const [countryLayer, setCountryLayer] = useState(null);
   const [borderGraphic, setBorderGraphic] = useState(null);
-  const [countryPinMarker, setCountryPinMarker] = useState(null);
 
   //Create the graphics layer on mount
   useEffect(() => {
     loadModules(["esri/Graphic","esri/layers/GraphicsLayer"]).then(([Graphic, GraphicsLayer]) => {
         const _borderGraphic = createGraphic(Graphic, gridCellDefaultStyles);
-        const _countryPinMarker = new Graphic({
-          symbol: simplePictureMarker(mapPinIcon, { height: 24, width: 24, yoffset: -10 }),
-          visible: !isCountryMode
-        });
-        const graphicsLayer = createGraphicLayer(GraphicsLayer, [_borderGraphic, _countryPinMarker], GRAPHIC_LAYER);
+        const graphicsLayer = createGraphicLayer(GraphicsLayer, [_borderGraphic], GRAPHIC_LAYER);
         setBorderGraphic(_borderGraphic);
-        setCountryPinMarker(_countryPinMarker);
         view.map.add(graphicsLayer);
       })
   }, [])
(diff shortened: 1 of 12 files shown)


aabdaab:
Hey 👋I have some comments about the flow itself. Love it or leave it.
* It's not very clear that you need to zoom a bit, wait for the country labels to appear and then click on the label to highlight a country. Wouldn't it be more transparent to the user to have a global political boundaries layer that you could toggle on and off? And when that layer is on, then the countries are clickable? In that case, you could lose the middle step of showing the `Explore this country box` and go straight to the local scene. It's also inconsistent with the other labels' behavior because if you go into landscape mode, the labels are not clickable there.
* This makes me think of another use case, what if they'd want to turn on other layers on the local scene? Now we have to go back a step, leave the local scene mode and turn on the layers and come back to the local scene to see them there.
* The panning and moving around on the local scene sometimes locks me out of the local scene :P and I'm left with the night sky 🌃without being able to go back. But you already flagged this issue to be tackled later.
* Somehow, I managed to arrive to this screen:
image
* The search bar isn't working on the local scene
HalfEarthExit

kfilip94:
Hey @fannycc, I'll try to remove this tilt transition on the local scene (from 0 to 40 degrees).

kfilip94:
So @aabdaab, answering your questions:
> It's not very clear that you need to zoom a bit, wait for the country labels to appear and then click on the label to highlight a country. Wouldn't it be more transparent to the user to have a global political boundaries layer that you could toggle on and off? And when that layer is on, then the countries are clickable? In that case, you could lose the middle step of showing the Explore this country box and go straight to the local scene. It's also inconsistent with the other labels' behavior because if you go into landscape mode, the labels are not clickable there.

We already spotted that with Fanny, she updated the design so we will make `Explore this country` bigger for now, but I agree with you, it's not very pleasant to use. Maybe @fannycc can give more details here.
image

***

> This makes me think of another use case, what if they'd want to turn on other layers on the local scene? Now we have to go back a step, leave the local scene mode and turn on the layers and come back to the local scene to see them there.

I think we're going to add the layers sidebar to the local scene, according to the design, right @fannycc @weberjavi ? 

***

> The panning and moving around on the local scene sometimes locks me out of the local scene :P and I'm left with the night sky night_with_starswithout being able to go back. But you already flagged this issue to be tackled later.

This bothers me as well, I didn't find the way to deal with that. The only option that I see now is to disable panning completely, but it means we can't grab & move that map, only zoom in/out. I'll check the arcgis documentation, maybe I'll find something helpful

***
> Somehow, I managed to arrive to this screen:

No idea, I'll try to reproduce & fix it

***

> The search bar isn't working on the local scene:

Since we have the country dropdown on the design, we don't really need this search `FIND PLACES` on the local scene - @fannycc, should I remove it from the local view?

fannycc:
Regarding this: 

> Since we have the country dropdown on the design, we don't really need this search FIND PLACES on the local scene - @fannycc, should I remove it from the local view?

Yes, this component will be consistent across all the experience, like changing mode or the map control buttons. If the user want to go to a different admin level they will need to use it

fannycc:
and about the rest of the comments: 
-The other cards to map biodiversity, protection and human encroachment should be present in the local view. 
-And the idea is to be able to turn labels on/off and use also the search bar as an entry point for the local view. But because of time constraints we decided to implement that in the following iterations. 
-regarding the card that triggers the country view, I agree, even if we made it bigger maybe we need to find a different solution in the next iterations if it's not clear yet. But we do need a middle step between clicking on the label and launching the analysis.

kfilip94:
Responding to @fannycc's  comments above, I created two Pivotal tasks to separate these modifications, and don't block this PR:
- [Add layers sidebar to the local scene view](https://www.pivotaltracker.com/story/show/172287970)
- [Fix the search widget on the local scene](https://www.pivotaltracker.com/story/show/172285694)

Another thing:
> The panning and moving around on the local scene sometimes locks me out of the local scene :P and I'm left with the night sky night_with_starswithout being able to go back. But you already flagged this issue to be tackled later.

To be honest I don't know how to deal with that. Apparently it's not possible to set min and max `altitude` for the local scene, according to the documentation:
image

Also, I haven't found information about how to constrain panning, I can only disable that but it's not an ideal behavior either. Without disabling it we can move the local scene outside the screen, so don't know which option serves better? @weberjavi @fannycc

weberjavi:
👁️ 🐽 👁️ 
 checking

kfilip94:
All code suggestions have been applied, thanks @weberjavi, they were super helpful! 
I wasn't able to fix the `Cannot read property '0' of null`, it seems to be related to `viewLocal.clippingArea` change. Also, I spotted another bug with multiple map markers, and we need to decide something about panning behavior. I'll update pivotal with those tickets.

So @weberjavi, ready again!

PR #182, opened 2020-04-14
Feature/fix metadata modal titles

kfilip94:
This PR solves three issues related to the metadata modal:
[PIVOTAL](https://www.pivotaltracker.com/story/show/172289471)

- Enables parsing the markup syntax in the metadata modal footer (you can test it by opening the `Community-based` layer modal):
**BEFORE**
image
**AFTER**
image

- Now we can pass custom metadata modal title to the checkbox component by specifying `metadataTitle` field. If the value for this field is not present, the value from the `name` field is set as a modal title:
image
this solves two problems that has been described in the pivotal ticket:
  - Titles on the metadata modals from the sidebar and the legend don't match,
_bug:_
image
  - Opening the metadata modal from the landscape sidebar copies the percentage value to the title, both protected areas and the human pressures widget,
_bug:_
image

BTW I noticed that tooltips in HE seem to be broken, they display in a weird way:
image
[Here](https://www.pivotaltracker.com/story/show/172315143) is the pivotal ticket for this issue.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "react-copy-to-clipboard": "^5.0.1",
     "react-dom": "^16.8.6",
     "react-hot-loader": "^4.8.2",
-    "react-markdown": "4.0.8",
+    "react-markdown": "^4.3.1",
     "react-redux": "^6.0.1",
     "react-responsive": "^8.0.1",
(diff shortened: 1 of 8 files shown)


weberjavi:
Checking ☝️

PR #183, opened 2020-04-15
Feature/countries dropdown

weberjavi:
### Description
This PR adds the chance to select a different country through a dropdown on the selected country data card.
It also adds the vertebrate species number to the data displayed on the country card.
### Designs.
Check the [inVision designs](https://projects.invisionapp.com/d/main#/console/17484384/409039744/preview).
Some changes to the designs have been made that are not reflected on the prototype:
  - Selected country name is not displayed on the dropdown.
### Testing instructions
Select a country and change the selection through the dropdown that can be toggled clicking on the arrow next to the country name.
### Feature relevant tickets
[Dropdown ticket](https://www.pivotaltracker.com/story/show/171812911)
[Endemic vertebrates ticket](https://www.pivotaltracker.com/story/show/172124765)

### DIFF
--- a/src/components/country-data-card/country-data-card.js
+++ b/src/components/country-data-card/country-data-card.js
@@ -1,15 +1,16 @@
 import React, { useEffect, useState } from 'react';
 import { connect } from 'react-redux';
 import { loadModules } from 'esri-loader';
-import { COUNTRIES_DATA_FEATURE_LAYER } from 'constants/layers-slugs';
+import { COUNTRIES_DATA_FEATURE_LAYER, COUNTRIES_LABELS_FEATURE_LAYER } from 'constants/layers-slugs';
 import mapStateToProps from './country-data-card-selectors';
 import { LAYERS_URLS } from 'constants/layers-urls';
 import { LOCAL_SCENE, GLOBAL_SCENE } from 'constants/view-props';
 import * as urlActions from 'actions/url-actions';
 import countryDataActions from 'redux_modules/country-data';
+import countriesListActions from 'redux_modules/countries-list';
 import Component from './country-data-card-component';
 
-const actions = { ...urlActions, ...countryDataActions};
+const actions = { ...urlActions, ...countryDataActions, ...countriesListActions };
 
 const CountryDataCardContainer = (props) => {
   const {
@@ -19,11 +20,15 @@ const CountryDataCardContainer = (props) => {
     setCountryDataLoading,
     setCountryDataReady,
     setCountryDataError,
+    setCountriesListLoading,
+    setCountriesListReady,
+    setCountriesListError,
     countryData
   } = props;
 
   const [countriesDataLayer, setCountriesDataLayer] = useState(null);
-
+  const [countriesListLayer, setCountriesListLayer] = useState(null);
+  
   useEffect(() => {
     loadModules(["esri/layers/FeatureLayer"]).then(([FeatureLayer]) => {
       const _countriesDataLayer = new FeatureLayer({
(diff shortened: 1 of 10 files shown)


weberjavi:
@kfilip94 something I forgot to mention on the PR description, but worth take into account when reviewing, is that I've created a `countries-list` redux module that is holding a `countriesList` array with the names and ISO code of all the countries of all the countries and also a `countriesData` array with an object holding the countries attributes (currently just the name and the iso code :). This might seem unnecessary because we are duplicating data just changing the data structures.
The rationale behind this is that in next iterations (when the layers on ArcGIS online are updated) we will hold countries data (regarding population, area...) on the `COUNTRIES_LABELS_FEATURE_LAYER` (we would need to rename it) because since it only has the centroids and not the complex borders polygons (that is the case with `COUNTRIES_DATA_FEATURE_LAYER`) is a more performant query to the server.
So in the near future we would need to refactor the `countries-list` and `country-data` redux modules to merge them into one and get rid of the `COUNTRIES_DATA_FEATURE_LAYER` request to gather the data.

weberjavi:
@kfilip94 let me review this PR before you get into it. The recently merged country mask is not working. Will let you know when it is ready again

weberjavi:
Done @kfilip94, ready again

kfilip94:
- Why we don't make the entire country label clickable? It's very weird being able to click only on the arrow.

- Maybe you should filter out the selected country from the countries list? Right now we can't click on that, it doesn't do anything, so to me makes sense to remove it or to set the font style to bold.
image

- There is no border-radius applied to the right corners of the dropdown:
image

src/components/fixed-header/fixed-header-styles.module.scss
```diff
@@ -97,6 +98,49 @@ $icon-size: 25px;
...
+  @include animationFunction($propertyToAnimate: all);
+  &.openList{
+    transform: rotate(270deg);
+  }
+}
+.dropdownWrapper {
+  position: relative;
+}
+
+.titleOptionsList {
```
kfilip94, line 116: I think you can add some styles to the vertical scrollbar. They don't look exactly like those on the design so you can adjust the color and padding if you want:
```suggestion
.titleOptionsList {
  @extend %verticalScrollbarStyled;
```

kfilip94:
The dropdown list doesn't display on mobiles, it's hidden somewhere under the sidebar:
image

src/components/country-data-card/country-data-card.js
```diff
@@ -34,6 +40,31 @@ const CountryDataCardContainer = (props) => {
     });
   }, [])
 
+  useEffect(() => {
+    loadModules(["esri/layers/FeatureLayer"]).then(([FeatureLayer]) => {
```
kfilip94, line 44: Do you think it would make sense to use the same hook for `countriesDataLayer` and `countriesDataLayer`? I get that using two hooks for two layers looks more readable, but we're loading `loadModules(["esri/layers/FeatureLayer"])` twice, am I understand that correctly?

src/components/fixed-header/fixed-header-styles.module.scss
weberjavi: 🙌 thanks

src/components/country-data-card/country-data-card.js
weberjavi: I would not update it right now since one of the hooks is gonna be removed once we have the data on the centroids layer, and for what I've seen on the `network` tab of the dev tools is not impacting (seems that the modules are being cached).

weberjavi:
@kfilip94 since we are going to handle mobile styles in another PR  this is ready again.

weberjavi:
I've created [this task](https://www.pivotaltracker.com/story/show/172375559) to tackle the clicking on the arrow issue. Fixed header is a convoluted component and is trying to satisfy multiple use cases at once. This turns out in conflicting functionalities for the same elements. Let's rethink this component.

weberjavi:
thanks

PR #186, opened 2020-04-20
API version upgrade

weberjavi:
### Description
This PR prepares the `master` branch to update the ArcGIS js API version from 4.13 to 4.15 on the production server.
So far the only breaking change that has been found is a small (and unfixed) bug when trying to `focus` the `Find place` `input` when opening the modal. A guard has been set to avoid the app crashing when trying to `focus` a not yet mounted DOM node, but I have not been able to fix it.
Remember to update the API version on the `.env` file to `4.15`
### Designs
_not aplicable_
### Testing instructions
Please test the whole app in case I missed something. 
### Feature relevant tickets
[API update](https://www.pivotaltracker.com/story/show/172406591) pivotal ticket

### DIFF
--- a/src/components/biodiversity-layers/biodiversity-layers.js
+++ b/src/components/biodiversity-layers/biodiversity-layers.js
@@ -19,7 +19,13 @@ const BiodiversityLayerContainer = props => {
       const target = new Extent({
         xmin, xmax, ymin, ymax
       })
-      view.goTo({target});
+      view.goTo({target})
+        .catch(function(error) {
+          // Avoid displaying console errors when transition is aborted by user interacions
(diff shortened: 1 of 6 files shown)


kfilip94:
This new API seems to be working better, I like this new transition effect when changing the order or opacity of the layers. The flickering on Firefox is gone, so that's a huge improvement!
The thing I spotted is that when quickly changing layers, entering/leaving the landscape mode, the console throws errors about aborted promises. It doesn't break the application but happens quite often:
image

weberjavi:
@kfilip94 ready again ☝️

weberjavi:
I'm gonna deploy this to production then 🚀

PR #187, opened 2020-04-20
Fix/national report cards bugs

kfilip94:
This PR fixes four issues that we found recently;

- [Several points glow when clicking on species widget](https://www.pivotaltracker.com/story/show/172405293)
For those species that don't have a `common-name`, the dataset sets the `NA` placeholder. The problem was that the species widget was using a `common-name` as a unique identifier, so once we select a species without a common name (`NA`), all other species with `NA` placeholder will highlight. The solution was to use the  `scientificName` field as an identifier.
_bug:_
Pasted Image at 2020_04_20_11_40 am

- [Provide scientific name when a common name is not available](https://www.pivotaltracker.com/story/show/172405463)
We replace `NA` placeholder for species without `common-name` with their scientific name:
image

- [Viewport clips the last option of the `human-pressures` radio group on the sidebar](https://www.pivotaltracker.com/story/show/172384011)
The problem which I spotted on every browser, at least on Windows, @weberjavi please check if on your machine it doesn't break anything:
_bug:_
Pasted Image at 2020_04_17_06_20 pm_big

- [Fix z-index of the search modal](https://www.pivotaltracker.com/story/show/172383916)
The problem appears since using the react portal solution for the local scene - the search widget is rendered in a different way than other widgets. We don't add it as a React component; instead, we add this as the esri widget to view, by using `view.ui.add()`. This means that even with the position `fixed` and greatest `z-index`, the search widget is displayed bellow other components, that are higher in DOM hierarchy. I fixed it by manipulating DOM and moving the search widget from the esri ui parent to the 'root' div, where other widgets exist. It's not a very reactive solution, but the quickest we can get I think.
_bug:_
Screenshot_1_big

### DIFF
--- a/src/components/data-global-sidebar/data-global-sidebar-styles.module.scss
+++ b/src/components/data-global-sidebar/data-global-sidebar-styles.module.scss
@@ -6,7 +6,7 @@
   @extend %verticalScrollBar;
   @include animationFunction($propertyToAnimate: all);
   position: absolute;
-  height: 100vh;
+  height: calc(100vh - #{$sidebar-top-margin});
   width: calc(#{$sidebar-width} + 15px);
   top: $sidebar-top-margin;
   left: $site-gutter;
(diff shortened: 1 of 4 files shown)


kfilip94:
Thanks! 👍
I added a link for this PR to each task in the `CODE` section, and also I added a comment to the pivotal ticket with the explanation of the search modal fix.

PR #192, opened 2020-05-15
Update marine grid layer url

weberjavi:
### Description
Read the title :point_up: 
### Designs
no :woman_artist: involved
### Testing instructions
Enter `landscapeMode` on coastal zones and check how both grids clip on the coast line
### Feature relevant tickets
https://www.pivotaltracker.com/story/show/172643383

PR #193, opened 2020-05-28
Feature/update protection data for disolved grid

weberjavi:
### Description
This PR updates the URL of the layer that gets queried to retrieve the data of the percentage of protected areas per grid cell. This update is needed due to the l[ast changes on the grid layer](https://github.com/Vizzuality/half-earth-v3/pull/192)

A small change on the legend item `height` has also been made to avoid the scroll bar to be displayed on a single item.
### Designs
No designs needed for this task
### Testing instructions
Go to a marine area and check if the percentage of protected areas, on the landscape sidebar widget, is accurate.
### Feature relevant tickets
https://www.pivotaltracker.com/n/projects/2321903/stories/172643383

### DIFF
--- a/src/components/legend/legend-styles.module.scss
+++ b/src/components/legend/legend-styles.module.scss
@@ -5,7 +5,7 @@
 
 $site-gutter: 20px;
 $map-attribution-height: 30px;
-$legend-item-height: 82px;
+$legend-item-height: 85px;
 $legend-width: 260px;
 
 %legend-ramp-text {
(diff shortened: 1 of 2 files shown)


kfilip94:
Hey, not sure if you've mentioned it somewhere, but I've noticed that the grid doesn't crop to the coastal lines, as it was[ in this PR](https://github.com/Vizzuality/half-earth-v3/pull/192). Now I see always rectangles, why is that?

weberjavi:
Yeah @kfilip94 sorry, you can check [this pivotal comment ](https://www.pivotaltracker.com/story/show/172643383/comments/214620602)(and thread) for the explanation. Basically the thing is that the marine grid data was not only clipped but changed, so the grids where not more overlapping and adjacent cells on coastal border had the same cell ID so we (@gretacv ) decided to disolve both grids to work with just one, so no more two grids.
Is a pity that we lose the clipped coastal border (it looked awesome) but it also was having some performance impact when we where to merge a bunch of coastal (marine and terrestrial) grid cells on a large an intricate piece coastal line. So in order to bring that back (in case we want to do it) we would need to simplify the feature layer to decrease the number of vertices of the features.

PR #201, opened 2020-07-27
Chore/separate webscenes for local and global view

kfilip94:
### Description
This PR introduces two separate Web Scenes for each global and local Scene View. This change is important for further local scene development, as we no longer need a connection between local and global scene views.
### Testing instructions
Go to the `Explore Data` and test the local scene map by trying different countries; just make sure that everything works as it supposes to.
### Feature relevant tickets
It needs to be merged before starting the [labels on the local scene task](https://half-earth-map.atlassian.net/browse/HE-44).

### DIFF
--- a/src/components/country-labels-layer/country-labels-layer-component.jsx
+++ b/src/components/country-labels-layer/country-labels-layer-component.jsx
@@ -81,7 +81,7 @@ const CountryLabelsLayerComponent = props => {
     return function cleanUp() {
       if (graphicLayer) { graphicLayer.remove(countryPinMarker)}
     }
-  }, [layerReady, countryExtent, countryISO])
+  }, [layerReady, countryExtent])
 
   useEffect(() => {
     if (layerReady) {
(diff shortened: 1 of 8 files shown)


src/components/country-labels-layer/country-labels-layer-component.jsx
```diff
@@ -63,7 +62,7 @@ const CountryLabelsLayerComponent = props => {
   useEffect(() => {
     let graphicLayer;
     let countryPinMarker;
-    if (layerReady && countryISO && !isCountryMode) {
+    if (layerReady && countryISO) {
```
weberjavi, line 65: Maybe not entirely related with this PR but can you please check the Pin adding/removing logic? It is not working correctly
image

kfilip94:
Hey @weberjavi, ready again, the last commit should fix the bug :crossed_fingers:

weberjavi:
Thanks @kfilip94 
Gonna merge this now :rocket:

PR #205, opened 2020-08-03
Chore/scene architecture update

weberjavi:
### Description
This PR updates the way different view are created to make it more explicit and less convoluted. 
Previous to this PR the way we handled having more than one scene in the same page by using the `doubleScene` component. This approach worked fine at first but is not really scalable, since the scenes share state, so reasoning what elements should be displayed in each of them was starting to be complex and a lot of state and guards should be taken into account to avoid adding bugs. Also because the scenes where built as one having separated initial states and configurations was hard to maintain.
I've added an `scenes` folder in order to create standalone scenes that can be then used on the pages as components. Each scene has it'sown config and initial state, this facilitates development experience, since they are standalone components decoupled from each other and also facilitates code splitting (scenes are dinamically imported so they are only fetched when the user visits that scene).
### Designs
This is a development chore task, so no designs needed.
### Testing instructions
Go to data page and play with it (enter country mode)
### Feature relevant tickets
scene architecture refactor: https://half-earth-map.atlassian.net/browse/HE-62
flags: https://half-earth-map.atlassian.net/browse/HE-54


weberjavi:
@fannycc if you don't mind I will merge this PR (since it's a chore dev task) and will make you and @gretacv reviewers for the functionality on the next PR with the updated functionality of the chart.
:ok_hand: ?

PR #210, opened 2020-08-18
Feature/country entry tooltip

weberjavi:
### Description
This PR adds a tooltip to the map when user selects a country label. This entry point for the national report cards substitutes the country card previously displayed on the sidebar.
### Designs
[InVision prototype](https://projects.invisionapp.com/d/main?origin=v7#/console/17484384/409039747/preview?scrollOffset=2558)
### Testing instructions
Click on a country label and play with the map (data globe)
### Feature relevant tickets
https://half-earth-map.atlassian.net/browse/HE-81


weberjavi:
The delay is on purpose. I've set the popup to display after the viewport centers the borders of the country. That way we avoid having the moving tooltip while we get into the country. It looked better.
Gonna try the borders update on hover

gretacv (approved):
Looks good to me, I have selected a country the pop up has appeared and then I have played a bit with the globe and the only _weird_ thing I have noticed is when the country disappears from the view by rotating. Like in the image. 
image

weberjavi:
@fannycc I've updated the code to highlight the hovered countries and the selected one. If you do not have any strong concern I will proceed with merging

fannycc:
Proceed, dear, proceed

PR #213, opened 2020-08-21
Feature/ranking chart

Bluesmile82:
SPI Ranking page

image

To improve:
- There might be some responsive problems from previous PRs
- Handle two-line countries @fannycc

image

### Designs
https://projects.invisionapp.com/d/main?origin=v7#/console/17484384/427356835

### Testing instructions

Try the sorting, clicking on the countries, ...

### Feature relevant tickets

https://half-earth-map.atlassian.net/browse/HE-123?focusedCommentId=10218&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel


fannycc (commented):
Design has been updated regarding some comments in this pull request: (long countries and scroll bar) https://projects.invisionapp.com/d/main?origin=v7#/console/17484384/427356835/preview?scrollOffset=4854

- Handle 2 lines countries. We won't handle them, since in desktop view there aren't many extra-long cases we'll add ellipsis before the number at the maximum. (example 12 in design, ignore the country Czech Republic isn't one of those long cases)

- Decrease space between number ranking and country name now is huuugeee

- Add scroll indicators when the user scrolls. (Not related with this PR but the sidebar should have one too) we could show them only when the user scrolls.

- The chevron next to the scroll indicator is facing right instead of bottom. 

- I can only see one sorting icon in the titles of species / human and protection. Where are the others? When user clicks in one of them they shouldn't change since there is no way to reset it. 

- When sorting countries by species, human or protection the numbers of the country should go accordingly with the desired ranking in my opinion. Now if I click in human i can read 228 - 3 - 45 - 60... It's a bit confusing. If the ranking of the countries have changed because of the user decision to sort by, the numbers should be consistent with that new order.

fannycc:
Below; screenshot of the single sorting icon and the chevron next to scroll ^
Screenshot 2020-08-24 at 11 23 41

weberjavi:
@fannycc some thoughts regarding your comments:
- I think some hint to the user is needed in order to let them know which sorting is selected. So if the user triggers an ascending sort on centered on Human encroachment data this should be evident some how. Maybe not making the rest of the arrows disappear but what do you think if I change the label (human) and the selected sorting arrow (ascendent/descendent) color to be white instead of grey?
- The title of the ranking reads `Ranking Species Protection Index` so on my understanding, despite the list sorting being dynamic, the ranking is always the same and refers to the SPI value. I think maintaining the ranking position allows the user to understand which is the variable that has more weight on the calculation of the countries SPI; but probably @gretacv or @tamaramegan can clarify this point.
-  Also, would it make sense to have some way to reset the sorting?

fannycc:
- Ok about the indication in white (I think mine was a bug because I couldn't interact with the rest of the filters after selecting one)
- Yeah I need more opinions here. For me is weird reading 223 - 34 - 67 but maybe it makes sense in order to make clear that a country with low SPI can be high in the list for other reasons. 
- I'm not sure how it works now the sorting (or how it should work). If I click in HUMAN the list is sorted by human pressures but what happen if I click again in HUMAN? isn't that going back to the original state? if so, then no reset should be needed

weberjavi:
So the sorting works in an ascending and descending order having into account the higher or lower percentages of the main values for each column (high human encroachment, protected area and endemic species). In that sense once you sort you can no go back to the original SPI ranking order (clicking again on the sorting arrows will rearrange the list by the ascending or descending order of the selected indicator)

fannycc:
so there isn't third position where i can make the indicator default. 1/ higher human encroachment 2/lower human encroachment 3/ default (first position)

weberjavi:
We can do it :)

fannycc:
I was just checking components from libraries to sort and filter etc and I couldn't find any way to reset. Are we inventing the internet?

weberjavi:
yes we are. Finally you got it :tada:

weberjavi:
Updated all this things... gonna merge it now to put it on the demo :rocket:

PR #214, opened 2020-08-25
Feature/similar countries filter

weberjavi:
### Description
This PR adds the possibility for the user to filter the countries displayed on the chart to see only a subset of them related to a selected subject.
A bug related to the device pixel aspect ratio and chart resolution has been fixed (by setting the `autoDensity` to true).
### Designs
Check [the inVision prototype](https://projects.invisionapp.com/d/main?origin=v7#/console/17484384/428930165/preview)
### Testing instructions
Go into the National Report Card for a given country. Select the `challenges` view and play with the chart.
### Feature relevant tickets
https://half-earth-map.atlassian.net/browse/HE-125
https://half-earth-map.atlassian.net/browse/HE-124


gretacv (commented):
Should the list disappear when clicking outside of it? at the moment the plot is interactive and the list does not disappear. 
image
Also, Gross National Income is with only one "m". For some countries there is no such kind of data and it shouldn't be possible to select.
image

fannycc:
I agree with Greta regarding the first comment.

PR #223, opened 2020-09-21
Feature/he 134 settings panel

Bluesmile82:
Add settings widget with checkboxes to toggle layers and country boundaries

The layer flow has changed, now only the Arcgis manager component creates layers and adds it to the map.
Once we add an active layer to the list a hook will create this on the map. 
When the layer is removed from the active layers it will be set with visibility false, so it will still be there but hidden.
In this way, we can control the flow of creation and visibility filtering of the layers.

### Description

image

### Designs

https://projects.invisionapp.com/d/main?origin=v7#/console/17484384/429754496/inspect?scrollOffset=4729

### Testing instructions
These settings affect in a general way all the maps

### Feature relevant tickets
https://half-earth-map.atlassian.net/browse/HE-134


weberjavi:
@Bluesmile82 this PR should be open against https://github.com/Vizzuality/half-earth-v3/pull/196 :octocat:

weberjavi (changes requested):
I've left some comments regarding code style, please consider them and let me know your thoughts.

As for the conversation we had with @fannycc the `settings` should not be displayed on the featured globe. So `hideSettings` `prop` need to be added to the `widget` component. That would have an impact on the layout and positioning of the widgets, since they are absolutely positioned :/ 

Regarding functionality there are some bugs that need to be addressed:
- When getting into `landscapeMode` `LabelsLayer` is displayed even if we toggled them on global. This happens because we are dynamically adding the layer when changing mode and creating the layer `on mount` through `handleLayerCreation` function. This can be bypassed by importing and creating the layer right away, removing the `isLandscsapeMode && ...` guard.

- Also inside `CountryLabelsLayer` we are controlling layer visibility taking into account if `isLandscapeMode`, that piece of code (`useEffect` hook on line 58 of `country-labels-component`) should be removed in order to avoid the layer appearing when changing mode.

- Also since country and global are different scene instances the selection does not get persisted. Probably we would need to store this settings and pass them between scenes. Let me know if you want to talk about what could be the best way to do this.

Bluesmile82:
Rebased and ready for testing. We might want to revisit some of the layers like the feature layer (currently filtering the view)

weberjavi:
:eye: :pig_nose: :eye: 
checking it

weberjavi:
@Bluesmile82 please check this out :point_up: when you have some time

src/scenes/data-scene/data-scene-config.js
```diff
@@ -12,8 +14,10 @@ export default {
   globe: {
     activeLayers: [
       { title: GRAPHIC_LAYER },
+      { title: CITIES_LABELS_LAYER },
```
Bluesmile82, line 17: So we are always adding these layers from the beginning. Don't we need to add them in other scene configs? I actually think this could be improved in the future and style the layers in some other place. Maybe as these are ISO dependant will be a bit more difficult. But its weird to me that this component just styles these layers and is in the middle of the sceen.

src/components/arcgis-layer-manager/arcgis-layer-manager.js
```diff
@@ -1,35 +1,23 @@
...
-    if (activeLayers && activeLayers.length) {
-      activeLayers.forEach((layer) => {
-        const layerConfig = MOLLayersConfig[layer.title];
-        handleLayerCreation(layerConfig, map);
-      });
-    }
+    // Add not already created activeLayers to the map
+    addActiveLayersToScene(activeLayers, layersConfig, map);
+    // Display layers in the correct order
+    setLayerOrder(activeLayers, map);
```
Bluesmile82: I like it, very clean 🚿

src/utils/arcgis-layer-manager-utils.js
```diff
@@ -19,3 +20,18 @@ export const setOpacity = (layer, activeLayers) => {
   const l = activeLayers.find(o => o.title === layer.title);
```
Bluesmile82, line 22: Could we have a more meaningful variable name?

PR #224, opened 2020-09-22
He 164 share button

Bluesmile82:
Share button for country cards and refactor

### Description
image

Add a functioning share button to the country cards, split button and share modal. Use variants for the three different share buttons:
- Icon (default): Small icon
- shortText: just 'Share'
- shortText: 'Share this info' the one on the country card

### Designs
https://projects.invisionapp.com/d/main?origin=v7#/console/17484384/421887229/preview?scrollOffset=4325
### Testing instructions
Please try all share buttons on the app

### Feature relevant tickets
https://half-earth-map.atlassian.net/browse/HE-164

### DIFF
--- a/src/components/featured-map-card/featured-map-card-component.jsx
+++ b/src/components/featured-map-card/featured-map-card-component.jsx
@@ -1,11 +1,12 @@
-import React, { useEffect } from 'react';
+import React, { useEffect, useState } from 'react';
 import cx from 'classnames';
-import ShareModalButton from 'components/share-modal';
+import ShareModal from 'components/share-modal';
+import ShareModalButton from 'components/share-button';
 import styles from './featured-map-card-styles.module.scss'
 import animationStyles from 'styles/common-animations.module.scss';
 import { ReactComponent as ChevronIcon } from 'icons/arrow_right.svg';
 
-const FeaturesMapCardComponent = ({ 
+const FeaturesMapCardComponent = ({
   view,
   className,
   selectedSidebar,
@@ -18,7 +19,7 @@ const FeaturesMapCardComponent = ({
   handle
 }) => {
   const isOpen = selectedSidebar === 'featuredMapCard';
-  
+
   const isOnScreen = isOpen && !isLandscapeMode && !isFullscreenActive && !selectedFeaturedPlace;
 
   const handleClick = () => {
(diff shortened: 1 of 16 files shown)


src/components/share-button/share-button-component.jsx
```diff
@@ -0,0 +1,87 @@
...
+    'data-for': 'shareButtonId',
+    'data-place': 'right',
+    'data-effect':'solid',
+    'data-delay-show': 0
+
+  };
+
+  const handleOpenShareModal = () => {
+    const { openShareModalAnalyticsEvent, viewMode } = props;
+    console.log({ openShareModalAnalyticsEvent, viewMode });
```
weberjavi, line 23: No log please :robot:

src/components/share-button/share-button-component.jsx
```diff
@@ -0,0 +1,87 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { ReactComponent as ShareIcon } from 'icons/share.svg';
+import cx from 'classnames';
+import ReactTooltip from 'react-tooltip';
+
+import styles from './share-button-styles.module';
+
+const ShareButtonComponent = (props) => {
+  const { theme, variant = 'icon', setShareModalOpen } = props;
```
weberjavi, line 10: Would it make sense to allow the component to consume a `text` `prop` to display by the button instead of having fixed `variants`?

src/components/featured-map-card/featured-map-card-component.jsx
```diff
@@ -35,31 +36,47 @@ const FeaturesMapCardComponent = ({
...
-      }
-    </div>) || null
-  )
+              <ShareModalButton
+                theme={{ shareButton: styles.shareButton }}
+                setShareModalOpen={setShareModalOpen}
+              />
+              <ShareModal
+                isOpen={isShareModalOpen}
+                handleClose={() => setShareModalOpen(false)}
```
weberjavi, line 63: I think a better approach here would be to pass the setter function instead of an arrow function on the render. And then using it on the container of the `ShareModalComponent` passing there the `false` param. That way we would avoid some undesired re render that might happen since this arrow functions [would be created each time](https://reactjs.org/docs/faq-functions.html#arrow-function-in-render).

src/components/share-modal/share-modal-component.jsx
```diff
@@ -1,26 +1,24 @@
...
+const ShareModalComponent = ({ handleClose, isOpen, shareSocialMedia }) => {
   const [activeTab, setActiveTab] = useState(LINK);
   const [copied, setCopied] = useState({ [LINK]: false, [EMBED]: false });
   const [shortLink, setShortLink] = useState(null);
   const [shareUrl, setShareUrl] = useState(null);
 
   const isActiveTabLink = activeTab === LINK;
   const currentLocation = window.location.href;
-  
+
```
weberjavi, line 21: Sorry for bringing this here, not related to this PR, but we need to create the share link only when the modal `isOpen`. By doing it that way we would get two needed things:
- Create the `bit.ly` link just when the user explicitly interacts with the share functionality (savin url creation from the free account :)
- Share the actual state of the app in some cases where the user updates it after the modal has been mounted (that happens for example when entering the national report cards and the user visits the challenges chart, for example, the share url was already created and when it is shared it points to the map instead to the chart).

Doing this would do the trick :point_down: 

```js
useEffect(() => {
    if (isOpen) {
      getShortenUrl(currentLocation)
      .then(link => {
        setShortLink(link);
      })
    }
  }, [isOpen])
```

weberjavi (commented):
Thanks for splitting the modal and the button :bow: 
Please check the comments and let me know your thoughts

src/components/share-button/share-button-component.jsx
Bluesmile82: I thought about it. But the text, in this case, is still always linked to the variant. We need these three different variants as for example the icon is in different positions. Maybe we could use this text prop but the markdown is already different in each variant

Bluesmile82:
Ready again!

src/components/share-button/share-button-component.jsx
weberjavi: :ok_hand:  fair enough

weberjavi (approved):
Thanks for the changes :bow:

PR #225, opened 2020-09-22
Fix layout and double lines

Bluesmile82:
image

### Description
- Add ellipsis to long countries
- Fix header alignment
- Add a bit more widht to ranking widget

### Testing instructions
Try different screens. Check long countries

### Feature relevant tickets
https://half-earth-map.atlassian.net/browse/HE-161


fannycc (approved):
Looks good to me. I've been playing around clicking some countries and sorting. Something I've noticed and that we'll need to address soon is to highlight the country we select because there is lack of feedback when the user clicks on a different country. Will create a task for me to give you the styles in Jira :) all the rest looks neat!

PR #226, opened 2020-09-23
Feature/nrc entry points

weberjavi:
### Description
This PR changes the way the user can get into the National Report Cards. The means to get into NRC are:
- Through interaction with the **countries layer**. 
  -  The hovered country border should highlight and the cursor turn into pointer style to give a hint to the user that it can be clicked.
  -  Once the country is clicked the entry tooltip is displayed and the country border remains highlighted. The view is centered on the country geometry.
- Through the `find places` **search widget**:
  -  Some updates to the search sources of the widget have been made. Now the widget offers suggestions from two different sources: the country borders layer and the Arcgis Worl geolocator, the suggestions from the later have been filtered in order to display only relevant features.
  -  When the user selects a suggestion from the countries layer the country tooltip is displayed after the view is centered on the selected country geometry.

**Missing bits**:
When selecting country through the search widget the country border is supposed to be highlighted but this is not yet implemented. I'd like to have the settings PR merged with this one before that gets implemented because as for now the border highlight happens inside the `countries-borders-layer` so the turning it off from the settings might have some impact.

**Extra stuff**:
The location functionality has been added to the search widget (`LOCATE ME`). 
Eventually the location widget could be removed from the map widgets. There's something that'd need to be done though, right now the location widget is responsible of updating url center coordinates whenever the view position is updated, this logic needs to be removed from there and placed somewhere else (maybe on the scene component or on a dedicated `manager`).
There's [a task](https://half-earth-map.atlassian.net/browse/HE-72) already created to handle that.

### Designs
[Search widget redesign](https://projects.invisionapp.com/d/main?origin=v7#/console/17484384/363963854/preview?scrollOffset=1110)

### Testing instructions
Interact with country borders on the [data globe](http://localhost:3000/dataGlobe).
Get into the NRCs by searching a country on the search widget.

### Feature relevant tickets
https://half-earth-map.atlassian.net/browse/HE-139


Bluesmile82:
I can see a problem with the interaction. When you select a country and the tooltip opens, the map doesn't zoom or pan to the country so when we enter it the map suddenly goes away from sight. Also when we go back to global we are zoomed in in a strange location. Try to click on a country that is a bit far from your position.

Bluesmile82:
> I can see a problem with the interaction. When you select a country and the tooltip opens, the map doesn't zoom or pan to the country so when we enter it the map suddenly goes away from sight. Also when we go back to global we are zoomed in in a strange location. Try to click on a country that is a bit far from your position.

Ah wait, it is zooming and panning but is just slow sometimes. Maybe we should wait for it to do it to allow clicking on the button.

src/components/countries-borders-layer/countries-borders-layer.js
```diff
@@ -1,83 +1,86 @@
...
-      if (!countryTooltip || countryTooltip !== attributes.GID_0) {
-        view.goTo(graphic.geometry)
-        changeGlobe({countryTooltip: attributes.GID_0, countryName: attributes.NAME_0});
-      } 
-    } else if (countryTooltip) {
-      console.log('outside labels layer')
-      changeGlobe({countryTooltip: null});
-    }
+  const onClickHandler = bordersLayerFeatures => {
+    flyToGeometry(view, bordersLayerFeatures);
```
Bluesmile82, line 40: Maybe we can wait for this to succeed before toggling the country tooltip. Not sure why it was not working fast enough on my side

src/pages/data-globe/data-globe-component.jsx
```diff
@@ -25,12 +25,12 @@ const DataGlobeComponent = ({
...
   handleGlobeUpdating,
   countedActiveLayers,
+  sortRankingCategory,
   localSceneActiveTab,
   isBiodiversityActive,
   isLandscapeSidebarCollapsed,
   countryChallengesSelectedKey,
-  sortRankingCategory
-}) => (
+}) => console.log(sceneMode) || (
```
Bluesmile82, line 33: Kill the log, human 🤖

src/styles/override/esri-search-widget.scss
```diff
@@ -81,30 +75,67 @@
...
+        left: 30px;
+        color: $white;
+      }
+    }
+
     strong {
       color: $white;
     }
+
+    &::-webkit-scrollbar {
```
Bluesmile82, line 131: Is this something that we reuse?

Bluesmile82 (approved):
Nice job! The selection is much better now! I still had some problems with lags on zooming and also on the search 'locate me'. Check my comments for that

src/components/countries-borders-layer/countries-borders-layer.js
weberjavi: Yeah, it happens to me sometimes too. I'm not sure why it has this lag (it is an internal arcgis API function `view.goTo`). But yeah something to investigate in order to try to improve it

PR #228, opened 2020-09-23
Don't show image if its missing

Bluesmile82:
Don't show image on Species list if its missing

### Description
image
### Testing instructions
See description image
### Feature relevant tickets
https://half-earth-map.atlassian.net/browse/HE-162


weberjavi (approved):
:framed_picture:

PR #229, opened 2020-09-24
Feature/he 74 challenges info

Bluesmile82:
### Description
image

I had to use a state prop to only change one of the modals and a useEffect to remove the additional info when we change of tab.

### Designs
https://projects.invisionapp.com/d/main?origin=v7#/console/17484384/421888341/inspect?scrollOffset=4614
### Testing instructions
Check the modals. Open it with the '?' button next to the Challenges title
### Feature relevant tickets
https://half-earth-map.atlassian.net/browse/HE-74


weberjavi:
@Bluesmile82 I'm taking that one :point_up: to address the changes on the metadata service agreed with Greta

weberjavi:
@Bluesmile82 check the changes if you will :point_up:

src/components/modal-metadata/modal-metadata-component.jsx
```diff
@@ -51,7 +51,9 @@ const ModalMetadata = ({ isOpen, handleClose, loading, title, metadata, addition
                   )
               }
             </dl>
-            {additionalContent}
+            {metadata && metadata.hasAdditionalContent === "TRUE" && (
```
Bluesmile82, line 54: Maybe we could parse this value before. But it's just an improvement

src/components/country-challenges-chart/country-challenges-modal-additional/country-challenges-modal-additional-component.jsx
```diff
@@ -26,25 +26,6 @@ const CountryChallengesModalAdditional = () => {
       <div className={styles.countriesLegend}>
         {CONTINENTS.map((continent) => renderContinent(continent))}
       </div>
-      <div className={styles.sizesLegend}>
```
Bluesmile82: There might be some styles code that we need to remove too

weberjavi: :+1:

PR #231, opened 2020-09-29
Feature/country highlighted species

weberjavi:
### Description
The `highlighted-species-list` component has been created. It gets a set of endemic and rare species for each country. A `random` [from the service](https://eowilson.maps.arcgis.com/home/item.html?id=aa62e9946df34e4ba176827c8ebc1b4d) attribute is used in order to display a different subset each time the user visits a country.
Two new services have been created `mol` (to get data from [MoL API](https://api.mol.org/1.x/docs/?m=uploader_ext)) and `esri-feature-service` (to abstract the way we get data from arcgis online services that are not meant to be displayed on the map but fetched to get data). 
`esri-feature-service` has been made to substitute the flow of creating and querying a `featureLayer` by a more common service pattern, hiding Esri logic inside the service. 

**Known issues**: Some species do not have an associated image on MoL API data and some others are not returning data at all. The later means that sometimes the widget does not display four species. A SQL `whereClause` can be passed as a prop to filter the set of features returned by the service.

### Designs
https://projects.invisionapp.com/d/main?origin=v7#/console/17484384/421888088/preview?scrollOffset=4509

### Testing instructions
Visit a national report card and scroll down to the species section on the sidebar

### Feature relevant tickets
https://half-earth-map.atlassian.net/browse/HE-26
https://half-earth-map.atlassian.net/browse/HE-167

### DIFF
--- a/src/services/esri-feature-service.js
+++ b/src/services/esri-feature-service.js
@@ -1,6 +1,7 @@
 import { loadModules } from 'esri-loader';
+import { LOCAL_SPATIAL_REFERENCE } from 'constants/scenes-constants';
 
-function getFeatures({ url, whereClause = "", outFields = ["*"], returnGeometry = false, outSpatialReference = 102100 }) {
+function getFeatures({ url, whereClause = "", outFields = ["*"], returnGeometry = false, outSpatialReference = LOCAL_SPATIAL_REFERENCE }) {
   return new Promise((resolve, reject) => {
     loadModules(["esri/tasks/QueryTask", "esri/tasks/support/Query"]).then(([QueryTask, Query]) => {
       var queryTask = new QueryTask({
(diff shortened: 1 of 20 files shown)


src/services/esri-feature-service.js
```diff
@@ -0,0 +1,26 @@
+import { loadModules } from 'esri-loader';
+
+function getFeatures({ url, whereClause = "", outFields = ["*"], returnGeometry = false, outSpatialReference = 102100 }) {
```
Bluesmile82, line 3: Wasn't this in a constant already? `outSpatialReference`

src/services/esri-feature-service.js
```diff
@@ -0,0 +1,26 @@
...
+  return new Promise((resolve, reject) => {
```
Bluesmile82, line 4: Maybe, in the future, we could add SWR if the response is not cached

Bluesmile82 (approved):
Working great! Thanks! 🦅

src/services/esri-feature-service.js
weberjavi: You're right :+1: 
Gonna give it a quick try

PR #235, opened 2020-10-08
Filter indicators, filters and data by available data

Bluesmile82:
### Description
This PR fixes the problems with missing data for indicators on the challenges chart
- Indicators without data of a country are not shown (low part of the chart)
- If a country is selected within an indicator, other countries bubbles' are not shown if that indicator data is not available
- On the filter menu if an indicator data is not available for a country the option related to that indicator won't be displayed

Missing: The animation is repeating so a bit annoying right now. I'm not sure if this is a priority right now. Let me know.

image

### Testing instructions
Go to the national reports.
- Select Antartica: Only 3 indicators should show
- If you select other country and you select GNI indicator - ATA (Anctartica) should never show.
- With Anctratica selected you shouldn't see GNI related filter on the dropdown
Try also with similar cases
### Feature relevant tickets
https://half-earth-map.atlassian.net/browse/HE-188

### DIFF
--- a/src/components/country-challenges-chart/country-challenges-chart-selectors.js
+++ b/src/components/country-challenges-chart/country-challenges-chart-selectors.js
@@ -66,11 +66,12 @@ const getFilteredData = createSelector(
     if (!plotRawData) return null;
     if (!selectedFilter || selectedFilter === 'all') return plotRawData;
     const relatedCountries = selectedCountryRelations[selectedFilter];
+    const hasNotNullXValue = (country, selectedKey) => (
+      country.xAxisValues[selectedKey] ||
+      country.xAxisValues[selectedKey] === 0
+    );
     return plotRawData.filter(
-      (country) =>
-        (relatedCountries.includes(country.iso) &&
-          country.xAxisValues[selectedKey]) ||
-        country.xAxisValues[selectedKey] === 0
+      (country) => relatedCountries.includes(country.iso) && hasNotNullXValue(country)
     );
   }
 );
--- a/src/components/country-challenges-chart/country-challenges-chart-selectors.js
+++ b/src/components/country-challenges-chart/country-challenges-chart-selectors.js
@@ -3,6 +3,10 @@ import { CONTINENT_COLORS } from 'constants/country-mode-constants';
 import { getCountryChallengesSelectedFilter, getCountryISO } from 'pages/data-globe/data-globe-selectors';
 import { countryChallengesChartFormats } from 'utils/data-formatting-utils';
 import * as d3 from 'd3';
+import {
+  INDICATOR_LABELS,
+  CHALLENGES_RELATED_FILTERS_OPTIONS,
+} from 'constants/country-mode-constants';
 
 const selectCountriesData = ({ countryData }) => (countryData && countryData.data) || null;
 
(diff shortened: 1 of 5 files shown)


src/components/country-challenges-chart/country-challenges-chart-selectors.js
```diff
@@ -48,7 +63,36 @@ const getFilteredData = createSelector(
...
+
+const getFilteredAvailableData = createSelector(
+  [getFilteredData, getCountryChallengesSelectedKey],
+  (data, selectedKey) => {
+    if (!data || !selectedKey) return null;
+    return data.filter(
+      (d) => d.xAxisValues[selectedKey] || d.xAxisValues[selectedKey] === 0
+    );
+  }
+);
```
weberjavi, line 76: You don't need the computation inside this selector. If you use the `getFilteredData` as source for the `data` prop the buggy animation does not happens anymore. 
```suggestion

```

src/components/country-challenges-chart/country-challenges-chart-selectors.js
```diff
@@ -74,10 +118,12 @@ const getYAxisTicks = createSelector(
 
 
 const mapStateToProps = createStructuredSelector({
-  data: getFilteredData,
+  data: getFilteredAvailableData,
```
weberjavi, line 121:
```suggestion
  data: getFilteredData,
```

weberjavi (changes requested):
Please check the suggestions, the second filtering of the data is causing the buggy animation.

weberjavi:
btw @Bluesmile82 , I've seen on the data that a `filter continent` is available so if you add it to the constants it will be displayed along the others :tada: 

```js
export const FILTERS_DICTIONARY = {
  // ... the other filters slugs
  filter_continent: 'Same continent countries'
}

export const CHALLENGES_RELATED_FILTERS_OPTIONS = [
  // ... the other options
  {slug: 'filter_continent', name: FILTERS_DICTIONARY.filter_continent}
]
```

src/components/country-challenges-chart/country-challenges-chart-selectors.js
Bluesmile82: We need the getFilteredAvailableData to be able to remove the countries bubbles when a specific country is selected and they don`t have value for the indicator selected. See the second case on the task description and testing instructions

Bluesmile82:
This is updated, now the animation is better, but still, a little glitch is happening. I wonder if it was like this before

weberjavi:
> This is updated, now the animation is better, but still, a little glitch is happening. I wonder if it was like this before

Maybe you can compare with the demos site.
The animation is buggy when a new country is selected (that is because the new scene is being rendered and the chart also is being updated and they are compiting for GPU resources) but once the scene is fully rendered the animation should run quite smooth

Bluesmile82:
I can't really find why the animation still has a little jump. Maybe we can check it on Tuesday. The rest of the chart is working fine now. Scatterplot props are not changing after render

weberjavi:
@Bluesmile82 I've fixed the duplicated animation. Gonna merge this 
Thanks :raised_hands:

PR #239, opened 2020-10-14
Fix/miscellaneous bugfixing

weberjavi:
### Description
Some bugs have been fixed:
-  Not existent page text store causing a crash on minimap load (it happened after local scene sharing).
-  Update local scene scene id to avoid console error because of exceeded call stack limitation (probably due to calling twice all the tiles from the firefly layer).
-  Use the `view.takeScreenshot` method instead the `@media print` for the scene in order to create an imageUrl and make the pdf consistent between browsers (thanks @craigmmills for spotting that).
-  Enable species widget carrousel species update by passing selected species to the scene (we missed that one when changing from pages to scenes :lion:)

### Designs
no designs for code mistakes :robot: 
### Testing instructions
Test it allllllll... we are about to release the beast :crab: 
But yeah, at least test the pdf and the species widget (on landscape view)
### Feature relevant tickets
:tickets:

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -9,7 +9,7 @@
     />
     <meta name="theme-color" content="#000000" />
     <link rel="manifest" href="%PUBLIC_URL%/manifest.json" />
-    <link rel="stylesheet" href="https://js.arcgis.com/4.15/esri/css/main.css">
+    <link rel="stylesheet" href="https://js.arcgis.com/4.17/esri/css/main.css">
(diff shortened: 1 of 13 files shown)


Bluesmile82 (approved):
I can't see the image on Firefox. The rest looks nice to me 🐛 🔨 

image

weberjavi:
:thinking: 
it is working for me (my version triggers a dialog to allow the display of background images and colors)

image

PR #241, opened 2020-10-15
Always center on country extent when entering

Bluesmile82:
### Description
Delete the code that was relying on the zoom and center to set the map position when we enter the National cards. Now we just center the country extent.

This was happening also due to lags in the country center
### Testing instructions
- Select a country boundary
- Before clicking into the modal pan the globe a lot
- Click on the modal to get to the National Report
- You should see the country centered
### Feature relevant tickets
https://half-earth-map.atlassian.net/browse/HE-180

### DIFF
--- a/src/components/local-scene-view-manager/local-scene-view-manager.js
+++ b/src/components/local-scene-view-manager/local-scene-view-manager.js
@@ -2,21 +2,14 @@ import { useEffect } from 'react';
 
 const LocalSceneViewManager = ({
   view,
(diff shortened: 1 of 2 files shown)


weberjavi (approved):
Great!!
As I said, we might want to think on migrating all these manager components into hooks that can be invoked on each scene. Take some time to think on what do you think would be the best approach and we might want to update this pattern and document it :newspaper_roll:

PR #242, opened 2020-10-16
Chore/api version upgrade 4.17

weberjavi:
### Description
This PR bumps API version to 4.17.
On the update process one issue has been found. When entering national report cards from the globe an `AbortError` is triggered by the `view.goTo` function, that was avoiding the view to tilt as expected. I have worked around that triggering a second goTo from inside the `catch` (this error does not triggers when the scene is already mounted and travelling from country to country.)
### Designs
_No designs_
### Testing instructions
Test the full app.
### Feature relevant tickets
https://half-earth-map.atlassian.net/browse/HE-179

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -9,7 +9,7 @@
     />
     <meta name="theme-color" content="#000000" />
     <link rel="manifest" href="%PUBLIC_URL%/manifest.json" />
-    <link rel="stylesheet" href="https://js.arcgis.com/4.15/esri/css/main.css">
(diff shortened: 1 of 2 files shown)


weberjavi:
@Bluesmile82 remember to update also the API version number on your `.env` file.

Bluesmile82 (approved):
Looks nice to me. I can't find any more problems. Maybe we would have to find out why this error is triggering

PR #243, opened 2020-10-16
Hide landscape feature dots

Bluesmile82:
### Description
These dots weren't able to remove before reaching the frontend so we have to cast a new renderer just to not render them but render the labels.
image

### Testing instructions
Go to landscape mode. Only the blue label items should be missing the point.

### Feature relevant tickets
https://half-earth-map.atlassian.net/browse/HE-129

### DIFF
--- a/src/components/labels-layer/labels-layer.js
+++ b/src/components/labels-layer/labels-layer.js
@@ -4,8 +4,8 @@ import { loadModules } from 'esri-loader';
 import { findLayerInMap } from 'utils/layer-manager-utils';
 import * as urlActions from 'actions/url-actions';
 import { LANDSCAPE_LABELS_LAYERS } from 'constants/layers-groups';
+import { LANDSCAPE_FEATURES_LABELS_LAYER } from 'constants/layers-slugs';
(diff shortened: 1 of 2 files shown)


weberjavi:
It is working fine. But I have one question.
We are already styling these layers on the `labels-layer` component. Maybe we can move this logic there? From line `56` we could add the code you've written for the `customFunction`.
What do you think?

`components/labels-layer/labels-layer.js`
```js
layers.forEach(layer => {
  layer.opacity = 1;
  layer.labelsVisible = true;
  layer.labelingInfo = labelingInfo;
  if (countryISO) {
    layer.definitionExpression = `GID_0 = '${countryISO}'`
  }
  if (layer.title === LANDSCAPE_FEATURES_LABELS_LAYER) {
      // Hides the dots but keeps the landscape feature layers
      layer.renderer = {
        type: 'simple',
        symbol: {
          type: 'simple-marker',
          size: 0
      }
   };
 }
});
```

Bluesmile82:
You are right. Updated

weberjavi (approved):
Great, thanks!!

PR #249, opened 2021-01-28
Feature/cypress automation

weberjavi:
### Description
This PR sets up a github workflow aimed to run tests on each push to the repository.
Several community build actions are used in this workflow:
- [Checkout](https://github.com/actions/checkout) 
- [Setup-node](https://github.com/actions/setup-node)
- [Cypress github action](https://github.com/cypress-io/github-action)

A branch protection rule has been added to avoid `feature/*` branches to be able to be delivered without passing tests. 

### Testing instructions
Check the [github actions tab](https://github.com/Vizzuality/half-earth-v3/actions) on the repo and you'll see the workflow acting.


cypress/integration/general-navigation.spec.js
```diff
@@ -0,0 +1,10 @@
+describe('General navigation', () => {
+  it('User lands on Data Globe if no path is provided ', function () {
+    cy.visit('https://half-earth.vizzuality.com/', {
+      onBeforeLoad (win) {
+        delete win.navigator.__proto__.serviceWorker
```
Bluesmile82, line 5:
```suggestion
      onBeforeLoad (window) {
        delete window.navigator.__proto__.serviceWorker
```

Bluesmile82 (approved):
This looks great! Do you think we should also include the 'fix/' starting branches into the branch protection rule?

weberjavi:
Yeah, I agree with the suggestion. We would need to set up rules for all the types of branches (`chore`, `fix`, `release`...). It would be great to have and document some kind of agreed naming.

PR #254, opened 2021-02-25
Feature/biodiversity tabs

Bluesmile82:
### Description

This PR  uses the tabs component to switch between the different biodiversity layers. If we have matching layers they will switch to their corresponding ones. If we don't have any layer selected it will default to all layers.
 
The PR also adds the priority layers config

image

### Designs
https://projects.invisionapp.com/d/main?origin=v7#/console/21108730/445055728/preview?scrollOffset=293

### Testing instructions
Switch between the different layer types with different layers selected.
### Feature relevant tickets
https://half-earth-map.atlassian.net/browse/HE-265


weberjavi:
☝🏻 Checking ☝🏻 
     👁️ 🐽 👁️

weberjavi:
Hey @Bluesmile82 can you rebase or merge with `develop`? There are some conflicts here

Bluesmile82:
This is ready for review again with the new priority layers. I just realized that there is a little bug in the positioning of the info-tooltip. I will check that

Bluesmile82:
This bug seems to have been there for a while. I think we could address it on a different PR

src/constants/landscape-view-constants.js
```diff
@@ -1 +1,7 @@
-export const ZOOM_LEVEL_TRIGGER = 8;
\ No newline at end of file
+export const ZOOM_LEVEL_TRIGGER = 8;
+
+export const LAYER_VARIANTS = {
```
weberjavi, line 3: This `landscape-view-constants` is storing ``landscape view related data. Maybe it makes more sense to store it inside another dedicated to biodiversity layers?

src/scenes/country-scene/country-scene-config.js
```diff
@@ -49,6 +49,7 @@ export default {
     countryChallengesSelectedFilter: 'all',
     localSceneActiveTab: LOCAL_SCENE_TABS_SLUGS.OVERVIEW,
     countryChallengesSelectedKey: 'N_SPECIES',
+    biodiversityLayerVariant: 'priority',
```
weberjavi, line 52: Maybe you can `import` the string from the `LAYER_VARIANTS` constant.

src/components/biodiversity-sidebar-card/biodiversity-sidebar-card-component.jsx
```diff
@@ -5,40 +5,46 @@ import Tabs from 'components/tabs';
...
         title="mapping"
-        category={biodiversity}
-        counter={countedActiveLayers[biodiversity]}
+        category={LAYERS_CATEGORIES.BIODIVERSITY}
+        counter={countedActiveLayers[LAYERS_CATEGORIES.BIODIVERSITY]}
         handleBoxClick={handleBoxClick}
         isOpen={isOpen}
       />
       <div
         className={cx(styles.layersTogglesContainer, { [styles.open]: isOpen })}
```
weberjavi, line 33: The `open` styles need to be updated. They have a fixed `height` in order to allow the container to animate that property; but since the `CategoryBox` does not have the same content in all of the variants maybe this value needs to be updated acordingly (if not the box is too big when priority layers are active)

weberjavi (changes requested):
Please check the comments. There's a bug related to batchLayerToggle new implementation.

Do you feel like adding the new styles for the sidebar and the thumb nails in this PR or do you prefere doing it in another one?

src/utils/layer-manager-utils.js
```diff
@@ -4,25 +4,31 @@ import { loadModules } from 'esri-loader';
...
-  if (areActive) {
-    const updatedLayers = layerNamesArray.reduce((acc, title) => {
-      removeLayerAnalyticsEvent({ slug: title });
-      return [...acc.filter(l => l.title !== title)];
-    }, activeLayers);
-    callback({activeLayers: updatedLayers });
-  } else {
-    const layersToAdd = layerNamesArray.map(title => {
+// Toggles all the layers passed as ids on the first parameter
+export const batchLayerManagerToggle = (layerIdsToToggle, activeLayers, callback, category) => {
```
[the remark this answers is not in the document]
Bluesmile82: I think we could add the styles in a different PR. This one is big enough

I will check this. So the previous batchToggle was either adding or removing layers. Not both. In this sense, it was not exactly being faithful to its name

src/components/biodiversity-sidebar-card/biodiversity-sidebar-card-component.jsx
```diff
@@ -5,40 +5,46 @@ import Tabs from 'components/tabs';
...
-            },
-            {
-              title: 'Richness'
-            },
-            {
-              title: 'Rarity'
-            }
-          ]}
-          onClick={() => console.info('Remove me I am just here for demo')}
+          tabs={Object.values(LAYER_VARIANTS).map((variant) => ({
```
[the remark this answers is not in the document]
Bluesmile82: I think this was added somehow in a rebase 🤦 Ill remove this

src/utils/layer-manager-utils.js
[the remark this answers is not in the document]
weberjavi: It was toggling the layers passed as param. They were being displayed or removed deoending on their previous state 🤷🏻‍♂️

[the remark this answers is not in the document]
weberjavi: Maybe you're rigth and the name is missleading. The use case for this _batch toggle_ is for layers that need to be turned off or turned on all at once in the same action; this happens when a group of layers (i.e human pressures) are turned off from the legend. It also happens when community protected layers (that are two different layers) are turned on and off from the sidebar.
So yes it only turns on or off all the layers passed as param.
Maybe is better to call the function twice instead of complicating the logic inside of it?

Bluesmile82:
Thanks for the feedback! Ready again

weberjavi:
☝🏻 checking againg (pardon the delay)

src/components/radio-group/radio-group-component.jsx
```diff
@@ -1,113 +1,68 @@
...
             }
           }}
           text={option.name}
         />
-        {isOptionSelected && (
+        {option.selected && (
           <div className={styles.toggle}>
             <InfoIcon
               className={styles.icon}
               onClick={() => handleInfoClick(option, variant)}
```
weberjavi, line 50: It seems that the `variant` is not reaching the component. It states `undefined` on the metadata modal

image

src/components/legend/legend.js
```diff
@@ -29,7 +34,11 @@ const LegendContainer = props => {
   const handleRemoveLayer = (layer) => {
     const { activeLayers, removeLayerAnalyticsEvent, changeGlobe } = props;
     if (layer.legendConfig.groupedLayer) {
-      batchLayerManagerToggle(LEGEND_GROUPED_LAYERS_GROUPS[layer.title], activeLayers, changeGlobe);
+      const activeGroupedLayers = intersection(
+        activeLayers.map((l) => l.title),
+        LEGEND_GROUPED_LAYERS_GROUPS[layer.title]
+      );
+      batchToggleLayers(activeGroupedLayers, activeLayers, changeGlobe);
```
weberjavi, line 41: Maybe we can move this `intersection` logic into its own utils function?

in `utils/layer-manager-utils.js`

```javascript
      const getActiveLayersFromLayerGroup = (layerGroup, activeLayers) => (
        intersection(activeLayers.map((l) => l.title), layerGroup)
      );
```
in `components/legend`

```javascript  
      const activeGroupedLayers = getActiveLayersFromLayerGroup(LEGEND_GROUPED_LAYERS_GROUPS[layer.title], activeLayers);
      batchToggleLayers(activeGroupedLayers, activeLayers, changeGlobe);
```

weberjavi (commented):
THANKS @Bluesmile82, working great now 🌍 

Check the comments, there's an issue with the metadata modal that needs to be fixed before merging.

Other thatn that looking 👍🏻

Bluesmile82:
Ready again!

weberjavi (approved):

weberjavi:
🚀

PR #256, opened 2021-03-01
Feature/circular button

weberjavi:
### Description
A `button` component has been created. It has two different variants `rounded` and `compound`. They are currently being used on the local scene sidebars.

`rounded`

image

`compound`

image

### Designs
https://invis.io/FH108BHHVCQ4#/445081450_sidebar_Overview

### Testing instructions
Go to National report cards and check the sidebar
### Feature relevant tickets
https://half-earth-map.atlassian.net/browse/HE-264


Bluesmile82:
This bug seems to have been there for a while. I  think we could address it on a different PR

src/components/local-scene-sidebar/local-scene-sidebar-styles.module.scss
```diff
@@ -5,33 +5,22 @@
...
   border: $sidebar-border;
   border-radius: $sidebar-border-radius;
   position: absolute;
-  height: 95vh;
+  height: calc(95vh + #{$sidebar-cards-padding});
   width: $local-sidebar-width;
   top: $local-sidebar-top-margin;
   left: $site-gutter;
-  padding-bottom: $site-gutter;
+  padding: $site-gutter 0;
```
Bluesmile82, line 16: I think we could add a margin on the top of the sidebar like on the design so the button has some space to the top of the page

src/components/button/component.jsx
```diff
@@ -0,0 +1,37 @@
...
+  tooltipText
+}) => (
+  <button 
+    onClick={handleClick}
+    data-tip={tooltipText}
+    {...tooltipStyles}
+		className={cx(
+      className,
+      {
+        [styles.rounded]: type === 'rounded',
```
Bluesmile82, line 27: I like this separation between style/layout 👍   We used 'variant' in Trase. But type also works :) Maybe I would recommend to put the className variable before this if we want to override something in some weird case (we shouldn't as the className should be only used for layout)

Bluesmile82 (approved):
Great!

PR #261, opened 2021-03-09
Feature/ranking update

Bluesmile82:
### Description
On the country pages ranking:
- Add a search that scrolls the table until the searched country
- Remove the info button
- Add sort dropdown that sorts by several options
- Remove previous sort functionality
- Fix layout
- Improve tooltip timing
- Use constants for all indicator and category slugs

EXTRA: Add missing labels on biodiversity layers sidebar cards
NOTE: Still have to decide if the option 'Sort by species richness' should be added and how

### Designs
https://projects.invisionapp.com/d/main?origin=v7#/console/21108730/445055733/preview?scrollOffset=907### Testing 

instructions
Go to the ranking on the country page:
- Test the search functionality
- Test the sorting dropdown

### Feature relevant tickets
https://half-earth-map.atlassian.net/browse/HE-273


src/components/ranking-chart/ranking-chart-component.jsx
```diff
@@ -1,63 +1,115 @@
...
+  scrollPosition,
+  searchTerm,
 }) => {
   const [hasScrolled, changeHasScrolled] = useState(false);
   const tableRef = useRef();
+  useEffect(() => {
+    if (scrollPosition > -1 && tableRef.current) {
+      const foundElement = document.getElementById(`country-${scrollPosition}`)
+      if (foundElement) {
+        foundElement.scrollIntoView();
```
weberjavi, line 88: I'm having some issues, in all the browsers I'm testing (firefox, chrome and vivaldi), when `scrollIntoView` is triggered. The whole container of the ranking view is scrolled and some gap appears at the bottom of the page. Maybe this is due to the positioning of the various `div`s.

image

src/components/ranking-chart/ranking-chart.js
```diff
@@ -14,35 +11,27 @@ import { openInfoModalAnalyticsEvent } from 'actions/google-analytics-actions';
...
 
   const handleCountryClick = (countryISO, countryName) => {
     const { changeGlobe } = props;
     changeGlobe({ countryISO, countryName, zoom: null, center: null });
   };
 
+  const handleSearchChange = (event) => {
+    const { changeUI } = props;
+    const { value } = event.target;
+    changeUI({ rankingSearch: value });
```
weberjavi, line 27: Not sure if this belongs to the `UI` params in the URL since is some action that the user makes but I dont think it needs to be stored on the `store`. We would be able to acommplish the same functionality using internal component state (`useState`) saving `actions` dispatching.

src/components/ranking-chart/ranking-chart.js
```diff
@@ -14,35 +11,27 @@ import { openInfoModalAnalyticsEvent } from 'actions/google-analytics-actions';
...
+    const { changeUI } = props;
+    changeUI({ sortRankingCategory: selectedFilter })
   }
 
   const handleCountryClick = (countryISO, countryName) => {
     const { changeGlobe } = props;
     changeGlobe({ countryISO, countryName, zoom: null, center: null });
   };
 
+  const handleSearchChange = (event) => {
```
weberjavi, line 24: I'd say that we should `debounce` firing the search change function, to avoid triggering it on each key stroke

weberjavi (changes requested):
Alvaro please check the comments, there's an issue with the scroll into view function.

Bluesmile82:
Ready again! Thanks for the feedback. I didn't realize about the jump with the scroll. scrollIntoView may not be useful at all

src/components/ranking-chart/ranking-chart.js
```diff
@@ -2,7 +2,7 @@ import React, { useState, useEffect } from 'react';
 import { connect } from 'react-redux';
 import Component from './ranking-chart-component';
 import mapStateToProps from './ranking-chart-selectors';
-
+import useDebounce from 'hooks/use-debounce';
```
weberjavi, line 5: We are already using `lodash`'s `debounce` on `countries-borders-layer.js`. Since we are already importing it maybe this `hook` is redundant (`lodash`'s `debounce` has some builtin options like `leading` and `trailing` that are used on the countries borders events so the hook is not ready to be used there).
One thing that might be needed is importing it as `import debounce from 'lodash/debounce'` instead of `import { debounce } from 'lodash'` as we are currently doing.
What do you think?

src/components/ranking-chart/ranking-chart-component.jsx
```diff
@@ -76,20 +21,18 @@ const RankingChart = ({
...
   useEffect(() => {
-    if (scrollPosition > -1 && tableRef.current) {
-      const foundElement = document.getElementById(`country-${scrollPosition}`)
-      if (foundElement) {
-        foundElement.scrollIntoView();
-        tableRef.current.scrollBy(0, -10);
-      }
+    const ROW_HEIGHT = 24;
+    if (scrollIndex > -1 && tableRef.current) {
+      tableRef.current.scroll(0, ROW_HEIGHT * scrollIndex);
```
weberjavi, line 33: I'm still seeing the same scroll displacement (do you experience the same)? Maybe is related with my browser version?

Bluesmile82: I can' reproduce this problem anymore neither in Chrome or firefox 😕  Are you talking about that white space due to the parent div scrolling?

src/components/ranking-chart/ranking-chart.js
Bluesmile82: lodash debounce in this case it's not easy to use. We could find another workaround but the problem is that debounce just returns a function and cant accept arguments on its contained function

weberjavi (approved):

weberjavi:
@Bluesmile82 tested on a mac and it was working fine. Maybe something related with my environment? Since you can not reproduce it lets merge this and I will open a bug with a proper description.
Thanks!!

Bluesmile82:
I changed this to use the lodash function but It keeps me thinking that it was just adding more complexity to the code. Take a look, maybe we can revert the last commit

weberjavi:
Sorry @Bluesmile82 this last update went under my radar 📡 😓 
Yeah you're right, feel free to revert the lodash `debounce` changes and merge this.
Thanks!!!

PR #264, opened 2021-03-22
Feature/feature map card update

Bluesmile82:
### Description
Some visual changes to the featured place card:
- Add position and number of the featured list
- Change title style, make it a link, add the position number
- Add close button and remove the back text
- Fix paddings

TODO: We have to move the 'Explore this area' button to the right and change the text to read the story. This should be done after removing the description text as there is not enough space there.

image

### Designs
https://projects.invisionapp.com/d/main#/console/21108730/445055724/inspect
### Testing instructions
Go to a featured place card

### Feature relevant tickets
https://half-earth-map.atlassian.net/browse/HE-263


Bluesmile82:
Updated with the mobile styles. Please check the fix crash commits. These happen mostly on reload. Maybe there is a better solution

PR #284, opened 2021-05-31
Docs initial structure

weberjavi:


weberjavi:
Please @gretacv let me know your thought regarding this first version of the documentation. Anything you think needs to be changed or restructured, any thing you're missing, anything you dislike...
🙏🏼 

You should be able to check them online here:
https://vizzuality.github.io/half-earth-v3/

gretacv (approved):
Javi, I think the main structure is great it is wonderful to have it divided depending on the "user". 
Maybe it is my personal preference but I would rather the external links open in a new tab. I think this is not possible with markdown, but it could be possible using html or maybe there is a config in mydocs that allow this. Or you might just want to ignore this comment. ☺️

weberjavi:
I agree with your point. Will address it 🙇🏼‍♂️

PR #285, opened 2021-06-04
Fix/safari misplaced chart labels

weberjavi:
### Description
Safari has a known bug related with `foreignObject` element translations on webkit based browsers. It is avoiding the labels inside the bubbles on the chart to be positioned properly (they are being animated as expected on other browsers). 
We have been forced to remove animation on the labels (placed inside a `foreignObject` inside the `svg` circles). I've tried to place the labels outside the `foreignObject` and position them as HTML elements but we where loosing the `z-index` reference of the `svg` objects and hence not being able to hide those that had overlapping bubbles.
This 'fix' just removes the animation of the labels (and updates its color to be the same as the background of the chart).
### Testing instructions
Visit the app with Safari browser. Inside a National Report Card go to the challenges view and play with the chart. Country iso labels should be properly placed inside the bubbles. 
### Feature relevant tickets
https://vizzuality.atlassian.net/browse/HE-30

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,5 @@
 npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
+
+/_site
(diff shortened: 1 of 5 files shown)


gretacv (approved):
chart labels are working correctly in the challenge plot.

PR #291, opened 2021-06-17
Feature/url params spike

weberjavi:
### Description
This PR creates a new National Report Card page. Previously we were displaying data and local scenes in the same page, leveraging URL `query` params (`countryISO`, `countryName`, `sceneMode`) to determine which scene and data to display without navigating away from the page. With this new approach more memorable urls are created for each national report card (now the country is defined in the URL `path` instead of in the `query`).
The needed router actions have been created, and navigation from and to the NRC page updated. With the new setup a link like `http://localhost:3000/nrc/KEN` will directly direct the user into Kenya's national report card.
It also updates the folder structure; all redux related code has been moved into the `store` folder.

The Half Earth Project white logo is also added to the interface in order to prepare the platform to link back to the project page once it is deployed on its own subdomain.
### Designs
[Logo prototype](https://projects.invisionapp.com/d/main?origin=v7#/console/21108730/453636707/preview?scrollOffset=2866)
### Testing instructions
`yarn && yarn install` and navigate the platform in order to test that all views are properly displayed.
### Feature relevant tickets
https://vizzuality.atlassian.net/browse/HE-36


gretacv (approved):
I navigated the global view, got into landscape view, switched the species in the widget, turned on and off layers and the url made a lot more sense.
And I went into a NRC and saw the short url ❤️ , I switched country by just changing the ISO3 in the url, worked perfectly. 
This is so exciting. :D 
Clicking the logo opened a new page, my only question there is if we should have a tooltip that explains what happens when you click on the logo. I appreciate that this might be something we figure out in user testing.

PR #342, opened 2021-12-10
Feature/update gp services to production

weberjavi:
### Description
This PR updates the GP services URLs in order to use the production server and the updated CRFs with the equal area projection.
The service code has been updated to use a JSON scaffold as a base to build the geometry that the server consumes. We then update the geometry `rings` with the ones provided by the user. 
### Testing instructions
Draw or update a shapefile and the analysis should be triggered and display accurate species data.
### Feature relevant tickets
https://vizzuality.atlassian.net/browse/HE-197


pablopareja:
@weberjavi I've simplified the URL definitions to make it less error prone ☝️ 
Let me know what you think 😉

weberjavi:
> @weberjavi I've simplified the URL definitions to make it less error prone
 Let me know what you think 

The thing is that those URLs are directly copy/pasted (through a button) from a field on ArcGis online item. So adding the variable in the string literal would mean to modify the copy pasted URL being then more error prone 😬

pablopareja:
Oh I see 🤔  I'll revert the commit then and merge the PR 🤷‍♂️ 😄

pablopareja (approved):

PR #353, opened 2021-12-20
HE-159 Protected areas table on AOI sidebar  (1st iteration: GADM-0 areas)

pablopareja:
### Feature relevant tickets
https://vizzuality.atlassian.net/browse/HE-159


tamaramegan:
@pablopareja the table is looking great! The only thing we have noticed is that the title says 'Protected Areas in' and does not have the country name.

PR #388, opened 2022-01-31
updated service urls and attribute for range

tamaramegan:
In this PR I updated the geoprocessing services for the AOIs. There is a modification in the field names because we are going to calculate both the percentage of a species relative to it global species range (per_global) and relative to its presence inside the aoi (per_aoi). We have also updated the attribute in the `src/utils/geo-processing-services.js` but check in case there are more places where we need to update this.

### DIFF
--- a/src/constants/analyze-areas-constants.js
+++ b/src/constants/analyze-areas-constants.js
@@ -61,7 +61,7 @@ export const SIDEBAR_CARDS_CONFIG = {
   },
   [PROTECTION_SLUG]: {
     title: 'What is already protected in this area?',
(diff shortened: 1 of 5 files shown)


tamaramegan:
Don't merge yet, I am getting an error from the Contextual layers that I need to check

tamaramegan:
Service paths are working in local. the only thing that needs to be changed is to eliminate multiplying by 100 the area protected of the WDPAs ( that is why we were getting values of 100% al the time)

tamaramegan:
Ready to merge

PR #480, opened 2022-06-09
Fix Firefox crashing on Challenges page

pedrogpimenta:
### Description
Firefox crashed because on the pulse animation 2 keyframes had the same `r` value (`1%`). I changed the second one to `1.1%`.
### Testing instructions
Open Firefox and navigate to, for example, https://map.half-earthproject.org/nrc/PER/challenges
### Feature relevant tickets
https://vizzuality.atlassian.net/browse/HE-404


Bluesmile82 (approved):
Thanks!

PR #481, opened 2022-06-13
Filter non-marine countries on Marine SPI ranking

pedrogpimenta:
### Description
When the Marine SPI filter is selected, it will check if the country has Marine value "True"
### Testing instructions
Open PR, navigate to NRC, select country (for example Brazil), and filter by Marine SPI: you should only see countries that have Marine SPI. 
### Feature relevant tickets
https://vizzuality.atlassian.net/browse/HE-431


Bluesmile82 (approved):
This could be even simplified to:

.filter((iso) => landMarineSelection === 'marine' || countriesData[iso].Marine === 'True')

But it's ok, I'm merging. Thanks!

PR #502, opened 2022-07-14
feat: format numbers by language[HE-467]

anamontiaga:
### Description
This PR includes the formation of those numbers that exceed or equal a thousand in relation to the chosen language in the app.

The XAxis ticks of the challenges chart still need to be translated. In order to read the locale when parsing the data, I would need to read it from the browser. So i would need to use lang set on [PR 501](https://github.com/Vizzuality/half-earth-v3/pull/501). 

### Feature relevant tickets
[HE-467](https://vizzuality.atlassian.net/browse/HE-467)


Bluesmile82 (approved):
Great job! Thanks

PR #527, opened 2022-08-19
fix: translate data strings on tooltip data [H-526]

anamontiaga:
### Feature relevant tickets
[HE-526](https://vizzuality.atlassian.net/browse/HE-526)


anamontiaga:
What about subnational names and protected area names that are not in the constants? 
Does they translate from Transifex? 
Why do we have some variables saved in constants?

Bluesmile82:
I opened a new PR due to the changes on the code https://github.com/Vizzuality/half-earth-v3/pull/534

PR #529, opened 2022-08-24
feat: separate map layers from analyse [HE-529][H-530]

anamontiaga:
### Description
- Separate layers and analyze tabs & add visual clues on layers and analyze tabs. Content changes will be made in separate tasks.
- Updated `framer-motion` to `v6.5.1`
- Because os `eslint` and `eslint import/order plugin` implementation, files should have a lot of changes. I have corrected the errors only of the files that I had to touch. My proposal is to correct them little by little so as not to generate conflicts, and then add husky.

- The layers from the data globe and AOI are passed through the UI url param activeCategoryLayers. Once they are reconciliated with the activeLayers the param is deleted.

### Designs
[Figma Prototype](https://www.figma.com/proto/2YAp3AS5lYCj2IWEAN8AJw/HE-Map---Work-in-Progress?node-id=4%3A19235&scaling=scale-down&page-id=0%3A1&starting-point-node-id=4%3A19661&show-proto-sidebar=1)

### Feature relevant tickets
[HE-529](https://vizzuality.atlassian.net/browse/HE-529)
[HE-530](https://vizzuality.atlassian.net/browse/HE-530)


src/store/selectors/aoi-selectors.js
```diff
@@ -1,14 +1,17 @@
 import { createSelector, createStructuredSelector } from 'reselect';
+
 import { selectGlobeUrlState } from 'selectors/location-selectors';
 
 export const selectAoisGeometries = ({ aoisGeometries }) => aoisGeometries.data;
 export const selectAoiId = ({ location }) => location.payload.id;
 export const selectLocationAreaType = ({ location }) => location.query && location.query.areaType;
 export const selectAreaTypeSelected = ({ aois }) => aois && aois.areaTypeSelected;
+export const setSidebarTabActive = ({ aois }) => aois && aois.sidebarTabActive;
```
anamontiaga, line 9: Not sure if this action should be here or opened a new store file

src/containers/sidebars/tabs-sidebar/component.jsx
```diff
@@ -0,0 +1,107 @@
...
+                    <div className={styles.layersIndicator}>
+                      {categoryActiveLayersCounter}
+                    </div>
+                    )}
+                    {displayMapLayersIcon(slug) && (
+                      <MapLayersIcon className={styles.tabIcon} />
+                    )}
+                    {displayAnalyzeAreasIcon(slug) && (
+                      <AnalyzeAreasIcon className={styles.tabIcon} />
+                      // TODO: Set indicator when some aoi is selected and user is on map-layers
```
anamontiaga, line 90: It is not yet implemented, I think it is necessary for the AOIs view to coexist with the data globe sidebar and to be able to add the layers in the AOI but I am not sure how to do it.

src/containers/sidebars/aoi-sidebar/aoi-sidebar-component.jsx
```diff
@@ -83,164 +98,176 @@ const AOISidebarComponent = ({
...
                 <Button
                   Icon={EditIcon}
                   type="icon-square"
                   handleClick={() => setIsEditingName(true)}
                   tooltipText={t('Edit area name')}
                 />
-              )}
-              {dataLoaded && (
+                )}
+                {/* // TODO: Share button will be place on side menu */}
```
anamontiaga, line 174: Temporarily disabled. It will be part of the new menu, developed in another PR.

src/containers/sidebars/tabs-sidebar/component.jsx
```diff
@@ -0,0 +1,107 @@
+import React, { useMemo } from 'react';
```
anamontiaga, line 1: There is a mismatch in the name we give to the files: some components contain the name of the same in each file, others contain only the dimension (index, selectors, etc). I prefer the second way of calling them because I think it is easier to search and less visually polluting, but we can decide which format is more interesting so that we can always use the same way.

src/constants/aois.js
```diff
@@ -20,6 +20,7 @@ export const BASE_LAYERS = [
   { title: 'gadm-0-admin-areas-feature-layer' },
   { title: 'mask-layer' },
   { title: 'specific-regions-tile' },
+  { title: 'admin_areas_feature_layer' },
```
anamontiaga, line 23: I am pretty sure is possible to improve this.

Bluesmile82: But this base layers should be the ones that are activeLayers on the aoi-scene/config.js. I think you can use them from there as they are exported

src/containers/sidebars/tabs-sidebar/component.jsx
Bluesmile82: For me is easier to search as you can search by the tabs-sidebar-component file directly

Bluesmile82:
image
Something is weird with this points on the precalculated AOIs it might have to do with the layers we load

Bluesmile82:
We have a bug: After selecting specific regions (clicking on one) and going back, and then selecting any other region (E.g. subnational) The app crashes as we still have areaType=specific_regions.

Bluesmile82:
I'm not sure if the Places for half earth future layer is loading. Can you see it?

PR #532, opened 2022-08-29
feat: new navigation menu implementation [H-531] [H-532] [HE-494]

anamontiaga:
### Designs

[Figma Prototype](https://www.figma.com/proto/2YAp3AS5lYCj2IWEAN8AJw/HE-Map---Work-in-Progress?node-id=272%3A63260&viewport=-256%2C1968%2C0.13&scaling=scale-down&starting-point-node-id=4%3A19661)

### Testing instructions
Please, add `REACT_APP_FEATURE_NEW_MENUS` feature flag to `.env`

Test if everything works fine with the feature disable. 

Once you set the feature, test:
- Feature is disable on mobile.
- Landing page should no longer have the side menu but a button that displays the "About the map" modal.
- All pages except the landing should have a right side menu with: 
       - location searcher
       - zoom controls
       - share button
       - help button which display a modal "How to navigate the map"
       - language switcher
- All main pages except the landing should have a globe menu indicator with the corresponding globe highlighted. (SEE NOTES ON USER STORY SAYING IT SHOULD BE BIGGER THAN IN THE DESIGNS)
- Both side and bottom menu should be disable on on boarding mode

### Feature user story ticket
[HE-494](https://vizzuality.atlassian.net/browse/HE-494)

### Feature relevant tickets
[HE-531](https://vizzuality.atlassian.net/browse/HE-531)
[HE-532](https://vizzuality.atlassian.net/browse/HE-532)

### Recommendations
We have two types of modals with two different overlays. The modal component is from the he-components library and you cannot conditionally override the overlay, so the component will be changed to a custom one in a future task.


src/containers/menus/globe-page-indicator/component.jsx
```diff
@@ -0,0 +1,59 @@
...
+} from 'router';
+
+import globeDiscover from 'images/globe-discover.png';
+import globeExplore from 'images/globe-explore.png';
+import globeNRC from 'images/globe-NRC.png';
+
+import styles from './styles.module';
+
+function GlobePageIndicator() {
+  const url = window !== undefined && window.location.pathname && window.location.pathname;
```
Bluesmile82, line 15: can we also get the path from a selector?

src/containers/menus/sidemenu/zoom-controls/index.js
```diff
@@ -0,0 +1,38 @@
...
+      const zoomWidget = new ZoomView({
+        view,
+      });
+      setZoomWidget(zoomWidget);
+      if (!hidden) {
+        view.ui.add(node, 'top-right');
+        // eslint-disable-next-line react/jsx-filename-extension
+        ReactDOM.render(<ZoomControlsComponent zoomWidget={zoomWidget} />, node);
+      }
+    }).catch((err) => console.error(err));
```
Bluesmile82, line 28: Ah, this is the reason why the zoom disappears sometimes. We could add a task with low priority to check if there is a better solution

src/utils/partners-utils.js
```diff
@@ -0,0 +1,202 @@
...
+import unEnvironmentLogo from 'logos/unEnvironment.png';
+import universityOfFloridaLogo from 'logos/universityOfFlorida.png';
+import vizzualityLogo from 'logos/vizzuality.png';
+import yaleLogo from 'logos/yale.png';
+
+export const getTexts = () => ({
+  partners: {
+    title: t('Half-Earth Project Map - Core Team'),
+    content: [
+      `<a style="text-decoration: underline; color: white" target="_blank" href="https://www.half-earthproject.org">The Half-Earth Project</a> ${t(' is an initiative of the E.O. Wilson Biodiversity Foundation. Map of Life utilizes geospatial species distribution data and analytics to guide where we have the best opportunity to conserve the most species. Vizzuality brings this information to life')}`,
```
Bluesmile82, line 31: We are missing t() on the texts inside the links. Just noticed

Bluesmile82 (approved):
Amazing job! Could you check the suggestion? Mostly bringing back that analytics and moving the url logic to selectors

src/containers/menus/sidemenu/zoom-controls/index.js
anamontiaga: going to add to JIRA

anamontiaga:
SearchLocation don't return any results since dev was merged

PR #544, opened 2022-09-12
fix: Feat/h 566 go back featured globe [H-566]

anamontiaga:
### Feature relevant tickets
[HE-566](https://vizzuality.atlassian.net/browse/HE-566)

### DIFF
--- a/src/containers/sidebars/featured-place-card/featured-place-card.js
+++ b/src/containers/sidebars/featured-place-card/featured-place-card.js
@@ -2,11 +2,11 @@ import { orderBy } from 'lodash';
 import React, { useState, useEffect } from 'react';
 import { connect } from 'react-redux';
 
-import { FEATURED } from 'router';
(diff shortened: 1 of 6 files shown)


src/containers/sidebars/featured-place-card/featured-place-card.js
```diff
@@ -27,10 +27,14 @@ function FeaturedPlaceCardContainer(props) {
     description: '',
   });
 
+  const handleClose = () => {
+    browsePage({ type: FEATURED });
+    window.location.reload();
```
anamontiaga, line 32: It can probably be solved in a more elegant way

Bluesmile82: Won't it be enough to reset the `selectedFeaturedPlace`? I don't think we have to browsePage or reload for this

PR #596, opened 2023-01-25
(fix): mobile devices

anamontiaga:
### Description
- Iphone SE & Samsung high resolution devices fixes.
- Fixed endemic cards on country card. Now marine data is displayed and endemic card is only shown when it has data.

### Feature relevant tickets
[HE-674](https://vizzuality.atlassian.net/browse/HE-674)

### DIFF
--- a/src/components/mobile-cards/mobile-cards-styles.module.scss
+++ b/src/components/mobile-cards/mobile-cards-styles.module.scss
@@ -3,9 +3,9 @@
 @import 'styles/ui.module';
 
 .container {
-  bottom: 4px;
+  bottom: 5svh;
   display: flex;
-  height: 40%;
+  height: 40svh;
   left: 3%;
   position: absolute;
   width: 94vw;
--- a/src/components/sidemenu-language-switcher/styles.module.scss
+++ b/src/components/sidemenu-language-switcher/styles.module.scss
@@ -20,8 +20,10 @@
     border: none;
     color: $white;
     font-weight: $font-weight-bold;
-    margin-top: 10px;
+    margin-top: 4px;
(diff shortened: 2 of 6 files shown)


src/components/search-location/component.jsx
```diff
@@ -87,9 +86,11 @@ function SearchLocation({
     waitingInteraction,
   });
 
-  useEffect(() => {
```
anamontiaga, line 90: For some reason the autofocus renderer is preventing the search result list from appearing. Subsequent renders are not working,

src/app-styles.module.scss
```diff
@@ -14,4 +14,5 @@
 
 .mobile {
   height: 100svh;
+  background-image: url('./assets/images/landing-bg.svg');
```
Bluesmile82, line 17: Why this?

anamontiaga: The background of the scene was missing in the landing but we don't have a scene in mobile so I put it in png.

anamontiaga:
@Bluesmile82 a new component has been added for the cards. despite the fact that the cyclical drag is blocked, a part of the first card is still displayed when the user is in the last one. Ideally, they couldn't drag in that direction.

package.json
```diff
@@ -28,6 +28,7 @@
     "eslint-plugin-cypress": "^2.12.1",
     "eslint-plugin-react": "^7.30.1",
     "esri-loader": "^3.7.0",
+    "framer": "^2.2.1",
```
Bluesmile82, line 31: You still have framer here!

Bluesmile82 (commented):
The tooltip has gone crazy 😢 
image

PR #607, opened 2023-02-01
(fix): zero cases on human modification indicator card 

anamontiaga:
### Description
We should have three different cases:
A. Country has very high and some human modification level data >> should display both.
B. Country has around 0% of very high human modification level data >> should display only some modification.
C. Country has around 0% of some human modification level data >> should display only very high modification.

### Testing instructions
A. Test with Portugal
B. Test with Namibia
C. Not found an example but added just in case

### Feature relevant tickets
[HE-693](https://vizzuality.atlassian.net/browse/HE-693)

[HE-693]: https://vizzuality.atlassian.net/browse/HE-693?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/containers/nrc-content/nrc-indicators/nrc-indicators-component.jsx
+++ b/src/containers/nrc-content/nrc-indicators/nrc-indicators-component.jsx
@@ -36,11 +36,7 @@ function Indicators({ countryData, landMarineSelection }) {
     hm_vh_mar,
     hm_ter,
     hm_mar,
-
-    hm_no_ter,
   } = countryData || {};
-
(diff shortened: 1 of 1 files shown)


Bluesmile82:
We need to address the protection indicator too. Please do the same for that one

Bluesmile82 (changes requested):
We need to address the protected indicator

PR #618, opened 2023-02-13
(fix): NRC feedback 

anamontiaga:
### Description
**Challenges chart:**

- [x] The dynamic axis labels that appear when you hover on a country circle overlap with the static min/max or variables labels on the axes.
- [x] If a point is the maximum value it doesn’t align with the x-axis label.

**Trend chart:**

- [x] In the SPI trend plot, the label of % protected area overlaps with SPI when they have the same value.

**Ranking List:**

- [x] When sorting the land or marine SPI from the lowest to highest and selecting a random country, the ranking list automatically switches back to the default sorting mode (from the highest to the lowest). 
- [x] If you have one country selected and change the sort order, it doesn’t track with the country you have selected, but rather stays on the same rank number. We think the view should track with the country selected.

### Testing instructions

- For Trend chart solution chosen solution has been change the legend, and set it outside the chart.

- To avoid overlapping issues on challenges chart it has been included a background on dynamic labels. Consider doing the same on the trend chart, since the labels are very close to the bottom

### Feature relevant tickets
[HE-711](https://vizzuality.atlassian.net/browse/HE-711)

[HE-711]: https://vizzuality.atlassian.net/browse/HE-711?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/charts/scatter-plot/scatter-plot-component.jsx
+++ b/src/components/charts/scatter-plot/scatter-plot-component.jsx
@@ -98,14 +98,9 @@ function ScatterPlot({
     if (!_data || !selectedKey) return null;
     const highValue = d3.max(_data, (d) => d.xAxisValues[selectedKey]);
     const lowValue = d3.min(_data, (d) => d.xAxisValues[selectedKey]);
-
-    const margin = (highValue - lowValue) / 20;
     // eslint-disable-next-line no-underscore-dangle
     const _formatFunction = countryChallengesChartFormats[selectedKey];
-    return [
-      _formatFunction(lowValue - margin),
-      _formatFunction(highValue + margin),
-    ];
+    return [_formatFunction(lowValue), _formatFunction(highValue)];
   };
 
   const xAxisTicks = getXAxisTicks(data, countryChallengesSelectedKey);
--- a/src/components/charts/scatter-plot/scatter-plot-component.jsx
+++ b/src/components/charts/scatter-plot/scatter-plot-component.jsx
@@ -98,9 +98,14 @@ function ScatterPlot({
     if (!_data || !selectedKey) return null;
     const highValue = d3.max(_data, (d) => d.xAxisValues[selectedKey]);
     const lowValue = d3.min(_data, (d) => d.xAxisValues[selectedKey]);
+
+    const margin = (highValue - lowValue) / 20;
     // eslint-disable-next-line no-underscore-dangle
     const _formatFunction = countryChallengesChartFormats[selectedKey];
-    return [_formatFunction(lowValue), _formatFunction(highValue)];
+    return [
+      _formatFunction(lowValue - margin),
+      _formatFunction(highValue + margin),
+    ];
   };
 
   const xAxisTicks = getXAxisTicks(data, countryChallengesSelectedKey);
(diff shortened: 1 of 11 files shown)


src/components/charts/scatter-plot/scatter-plot-component.jsx
```diff
@@ -98,9 +98,14 @@ function ScatterPlot({
...
     const highValue = d3.max(_data, (d) => d.xAxisValues[selectedKey]);
     const lowValue = d3.min(_data, (d) => d.xAxisValues[selectedKey]);
+
+    const margin = (highValue - lowValue) / 20;
     // eslint-disable-next-line no-underscore-dangle
     const _formatFunction = countryChallengesChartFormats[selectedKey];
-    return [_formatFunction(lowValue), _formatFunction(highValue)];
+    return [
+      _formatFunction(lowValue - margin),
+      _formatFunction(highValue + margin),
```
Bluesmile82, line 107: I think this is probably causing values to go over 100%
image

http://localhost:3000/nrc/STP/challenges?ui=%7B%22countryChallengesSelectedFilter%22%3A%22filter_steward%22%2C%22countryChallengesSelectedKey%22%3A%22protection_needed_ter%22%7D

src/components/ranking-chart/ranking-chart-component.jsx
```diff
@@ -68,6 +68,22 @@ function RankingChart({
     }
   };
 
+  useEffect(() => {
```
Bluesmile82, line 71: This seems to only be working when we sort SPI and not species, human modification or protection. Not sure why

src/containers/nrc-content/nrc-trend/nrc-trend-component.jsx
```diff
@@ -35,6 +35,16 @@ function Trend({
             landMarineSPI={land ? t('land SPI') : t('marine SPI')}
           />
         </p>
+        <div className={styles.areaChartLegend}>
```
Bluesmile82, line 38: We need to add this also to the PDF:

image

Also, please remove the commented code if its not used anymore

src/components/ranking-chart/ranking-chart-component.jsx
Bluesmile82: Also, this solution is very similar to the selectedLandMarineOption useEffect above. Do you think we could extract some function to avoid repeating code? Also, maybe tableRef.current would need to be a dependency

Bluesmile82 (changes requested):
Please check the comments. Some adjustments need to be made

src/components/charts/scatter-plot/scatter-plot-component.jsx
anamontiaga: New solution founded:
Should we add it to yAxis? Is a little weird to have the 0 value up to the Axis.

Screenshot 2023-02-13 at 20 28 30

src/components/ranking-chart/ranking-chart-component.jsx
anamontiaga: I have add a `selectedCountryIndex` state to calculate dynamicaly the index of the country (it was only working with SPI because it was getting the id of the array and not the position in the sorted array)  but it doesn't work yet because it have a little delay.

anamontiaga: fixed

Bluesmile82 (approved):
Well done! Thanks

PR #619, opened 2023-02-13
Fix Aoi recover from DB

Bluesmile82:
### Description
The AOI sharing was not working correctly
### Testing instructions
1. Draw custom AOI
2. Copy AOI share URL
3. In a new window open that URL
The AOI should show
### Feature relevant tickets
https://vizzuality.atlassian.net/browse/HE-706?atlOrigin=eyJpIjoiYjdmYjU1ZTA4NGM2NGY0MDkyYmI5MWE2ZGMxNmM2MjEiLCJwIjoiaiJ9


anamontiaga (commented):
When paste AOI url, the AOI is shown but the drawn area is not. Is that right?

PR #620, opened 2023-02-14
Fix loading with exception

Bluesmile82:
### Description
This check is for the case where the user goes from other AOI to a WDPA and the old data is still loaded
### Testing instructions
Go to a country AOI page
Go to the protected areas modal
Click on a PA
The loading state should show until we have PA data instead of showing the country data
### Feature relevant tickets
https://vizzuality.atlassian.net/browse/HE-717?atlOrigin=eyJpIjoiZTZkOTZjODMwNmFjNDgzOWI0ZWMyZGJmOGI5MzdmNmYiLCJwIjoiaiJ9

### DIFF
--- a/src/containers/scenes/aoi-scene/index.js
+++ b/src/containers/scenes/aoi-scene/index.js
@@ -33,6 +33,9 @@ import mapStateToProps from './selectors';
 
 const actions = { ...urlActions, ...aoisGeometriesActions };
 
+const INITIAL_SPECIES_STATE = {
+  species: [],
+};
 // Protected areas are fetched on protected areas modal except for PA type AOIs
 function AOIScene(props) {
   const {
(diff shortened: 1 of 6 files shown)


anamontiaga (commented):
I keep seeing the country data

src/containers/scenes/aoi-scene/index.js
```diff
@@ -136,11 +136,20 @@ function AOIScene(props) {
 
   // Reconcile all data until completely loaded
   useEffect(() => {
+    const hasWDPAData =
+      contextualData && (contextualData.WDPAID || contextualData.DESIG_T);
```
anamontiaga, line 140: If we are going to introduce a condition maybe we can check with precalculatedSlug && country.iso.
If slug is wdpa but we have countryISO, this countryData should not appear.

Bluesmile82: But maybe we have the same problem with regions or other AOI. So I think its better to know if we have the PA data

PR #621, opened 2023-02-14
(fix): check landscape

anamontiaga:
### Description
It seems that the mediaqueries for tablet landscape and mobile-landscape work correctly, but there is a problem with the positioning of the country tooltip in NRC.
However, on the first render, it didn't quite fit right.

### Testing instructions
Go to NRC and select a country in map. Check how tooltip fits in landscape and portrait for tablet and mobile devices.

### Feature relevant tickets
[HE-695](https://vizzuality.atlassian.net/browse/HE-695)

[HE-695]: https://vizzuality.atlassian.net/browse/HE-695?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/containers/nrc-content/nrc-vertebrates/nrc-vertebrates-styles.module.scss
+++ b/src/containers/nrc-content/nrc-vertebrates/nrc-vertebrates-styles.module.scss
@@ -29,6 +29,7 @@
 
     @media #{$tablet-only} {
       grid-template-columns: 1fr 1fr;
+      row-gap: 10px;
     }
 
     .endemicCard {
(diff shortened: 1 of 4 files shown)


Bluesmile82:
Could you check this position of the tooltip on the Ipad air screen size? Is not the end of the world but its weird that is absolute and on the top 
image

Bluesmile82 (approved):
The arrow on tablet wont be exactly pointing to the country so lets skip it for now. Thanks!

PR #627, opened 2023-02-20
Fix/sps-chart-filter

Bluesmile82:
### Description
- When the dropdown filters the selected species -> filter also the chart and legend
- Fix the numbers on the legend when we drag the cursors
- Fix the initial height so it doesnt overflow
### Designs
_Link to the related design prototypes (if relevant)_
### Testing instructions
_Explain how the reviewer should test this PR (which URL should be joined, which steps should be performed...)_
### Feature relevant tickets
_Link to the related task manager tickets_


anamontiaga (commented):
Search input only works when user press ENTER after write letters to look for. Is that the correct behaviour?

Bluesmile82:
@anamontiaga Yes, that is the behaviour for now. We can check with Andreia at some point to change it

PR #630, opened 2023-02-22
Feature/ts-store

Bluesmile82:
## Start typing the store
changes
### Description
I started typing the store and did some changes but as we are migrating to redux/toolkit  its better to do it on a different PR

The problem here is that we need to type the thunk actions from the redux-tools library that is on vizz repo (which is definitely not ideal). Also the code splitting system will make it more complex. So the solution will be the migration to redux/toolkit
### Feature relevant tickets
https://vizzuality.atlassian.net/browse/HE-720?atlOrigin=eyJpIjoiMzA4ZGI2NGY4OGVkNGM5OWIwYzVhMzAyNjhiN2NhZmQiLCJwIjoiaiJ9


declarations/modules/redux-tools.d.ts
```diff
@@ -0,0 +1,243 @@
+// Type definitions for redux-actions 2.6
```
anamontiaga, line 1: Do we need this comment?

anamontiaga (approved):

declarations/modules/redux-tools.d.ts
Bluesmile82: Yes, let's leave it as a reference

PR #660, opened 2024-03-23
Fix hide of Click options layers

cmanciero:
### Description
Currently when switching from Click to another option and back, the last selected option is deselected. This fix will keep that option active when going back to Click tab
### Testing instructions
Access the Analyze Area tab for the Data globe. Choose "Search" selection type and then select "Click" selection type. You will noticed that the regions are still able to be highlighted and clicked on

### DIFF
--- a/src/containers/sidebars/data-global-sidebar/analyze-areas-sidebar-card/component.jsx
+++ b/src/containers/sidebars/data-global-sidebar/analyze-areas-sidebar-card/component.jsx
@@ -139,7 +139,11 @@ function AnalyzeAreasCardComponent({
                     id={option.slug}
                     option={{ ...option, name: option.label }}
                     checked={selectedOption?.slug === option.slug}
-                    onChange={() => handleOptionSelection(option)}
+                    onChange={() => {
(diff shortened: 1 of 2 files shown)


Bluesmile82:
There is still a problem as you noticed with the highlight on the layer when we clear the selection. This is because we didn't contemplate the option of this layer being removed. 

In the feature-highlight-layer component which is the responsible of painting the outline (its an empty component that renders null but adds this effect to the graphic layer) we set the graphics layer with the effects on click and hover and add them to the map. 

https://github.com/Vizzuality/half-earth-v3/blob/analyze-area-toggle/src/containers/layers/feature-highlight-layer/index.js#L40

But it seems that once we clean up we don't delete this graphics layer so when we reinit it we have the graphic layer but the border graphics are not set.

Could you check this and find a solution? Maybe we could delete this graphic layer or reassign the graphics to it.

cmanciero:
@Bluesmile82 I updated the code to add the GraphicLayer back if going from Clearing Selections and back to choosing a layer

Bluesmile82 (approved):

PR #670, opened 2024-04-10
Develop to main

cmanciero:
Merge develop to main


Bluesmile82 (commented):
Hi Chris! Could you check with the team if the WDPA data layer I added is ok to go to production? Sofía says is working fine but just to be sure.

Also, maybe is better to wait for the visible event of the layer instead of the loaded. As the layer loads but some data is still not there it doesn't seem to be visible immediately. 

https://community.esri.com/t5/arcgis-javascript-maps-sdk-questions/handle-listen-layer-visibility-events/m-p/1275690#M80761

johnWilshire (approved):

PR #686, opened 2024-06-10
Upgrade to node 20.5.0

Bluesmile82:
Vercel requires of a node version higher than 16. I upgraded it to 20.5.

Please check the app carefully for any bug caused by the change. My tests were all successful.


cmanciero (approved):
LGTM, most of the changes are additions of aria-labels

PR #705, opened 2024-12-02
Switching tabs, reset layers

cmanciero:
Remove selected layers on Data Globe when switching tabs

### Description
When switching between Map Layers and Analyze Areas, remove selected layer (reset) from the map

### Testing instructions
Select a layer or multiple in the Map Layers section then click on Analyze Areas tab, selected layers should be removed.
Select a radio option in Analyze Areas then click the Map Layers tab and see layers be removed.


Bluesmile82:
About the feature: The biodiversity layers have a radio button so the idea is that one layer is always selected. Shouldn't we select the all taxa layer back when we return to map layers?

Also when we go to analyse areas the click on the map select options are also part of a radio. Ideally the first option should be selected.

I'm not sure about the purpose of this change but it could be a bit distracting for the user lose the selection of layers that they have, also on the aoi pages when we come to the analysis tab. Maybe there is another option

PR #710, opened 2025-02-11
Develop

cmanciero:
### Description
Navigate to https://map.half-earthproject.org/aoi/8033?precalculatedLayerSlug=wdpa-oecm-feature-layer&OBJECTID=7013&ui=%7B%22aoiId%22%3A8033%7Dlayer&OBJECTID=7013&ui=%7B"aoiId"%3A8033%7D
And you will notice the application crashes. This is due to No species being returned for the Protected Area

The fix is to add a check before the Species Card component is loaded to validate that species are in the data before attempting to load the card

### Testing instructions
Validate page does not crash and continues to work


johnWilshire (approved):
I would have done it a little differently

PR #719, opened 2025-04-04
Develop

cmanciero:
More layer url changes

### DIFF
--- a/src/components/ranking-chart/ranking-chart-component.jsx
+++ b/src/components/ranking-chart/ranking-chart-component.jsx
@@ -55,7 +55,7 @@ function RankingChart({
   const tableRef = useRef();
 
   const scrollTableVertically = (distance) => {
-    tableRef.current.scroll(0, distance);
+    tableRef.current.scrollTo({top: distance});
   };
 
   useEffect(() => {
@@ -84,7 +84,7 @@ function RankingChart({
     const PADDING = 20;
     const selectedCountry = data.find((d) => d.iso === countryISO);
 
-    if (selectedCountry) {
+    if (selectedCountry && searchTerm.length === 0) {
       scrollTableVertically(ROW_HEIGHT * selectedCountryIndex - PADDING);
     }
   }, [handleSortClick, tableRef.current, selectedCountryIndex]);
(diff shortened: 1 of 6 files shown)


src/components/ranking-chart/ranking-chart-component.jsx
```diff
@@ -55,7 +55,7 @@ function RankingChart({
   const tableRef = useRef();
 
   const scrollTableVertically = (distance) => {
-    tableRef.current.scroll(0, distance);
+    tableRef.current.scrollTo({top: distance});
```
Bluesmile82, line 58: The table is not scrolling to the selected country on reload now. Could you please check this?

src/constants/layers-urls.js
```diff
@@ -201,9 +201,9 @@ export const LAYERS_URLS = {
 
   // Protected areas WDPA
   [PROTECTED_AREAS_VECTOR_TILE_LAYER]:
-    'https://tiles.arcgis.com/tiles/IkktFdUAcY3WrH25/arcgis/rest/services/ProtectedAreas_202504/VectorTileServer',
+    'https://tiles.arcgis.com/tiles/IkktFdUAcY3WrH25/arcgis/rest/services/ProtectedAreas_20250416/VectorTileServer',
```
Bluesmile82, line 204: Im not seeing the protected areas anymore in the map. Is it possible to test the changes on staging before doing this PR to production?

Bluesmile82 (changes requested):

src/constants/layers-urls.js
cmanciero: Protected areas are showing up fine here - https://half-earth-v3.vercel.app/dataGlobe. I noticed sometimes it doesn't show but then I refresh and it shows fine

Bluesmile82: Ok, I can see them after reloading. Maybe we will have to check why they don't show at first.

Another issue: This might be on purpose but the layers are cleared when I reload the page. Maybe this was required at some point

Bluesmile82 (approved):
Thanks for the changes

PR #722, opened 2025-04-21
Butterfly-revamp

cmanciero:
URL updates where pushed before data was fully vented and minor issues are being caused. This PR reverts those URL changes and also makes Butterflies data fully functional

### DIFF
--- a/src/containers/sidebars/data-global-sidebar/analyze-areas-sidebar-card/index.js
+++ b/src/containers/sidebars/data-global-sidebar/analyze-areas-sidebar-card/index.js
@@ -52,15 +52,13 @@ export const getWarningMessages = (t, locale) => ({
     description: (size) => (
       <span>
         {t(
-          'The maximum size for on the fly area analysis is {number} km{sup}.',
+          'The maximum size for on the fly area analysis is {number} km.',
           {
             number: getLocaleNumber(HIGHER_AREA_SIZE_LIMIT, locale),
-            sup: <sup>2</sup>,
           }
         )}{' '}
-        {t('The area that you are trying to analyze has {number} km{sup}.', {
+        {t('The area that you are trying to analyze has {number} km.', {
           number: getLocaleNumber(size, locale),
-          sup: <sup>2</sup>,
         })}{' '}
         {t('Please select a smaller area to trigger the analysis.')}
       </span>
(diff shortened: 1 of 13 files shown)


src/containers/sidebars/data-global-sidebar/analyze-areas-sidebar-card/index.js
```diff
@@ -52,15 +52,13 @@ export const getWarningMessages = (t, locale) => ({
...
         {t(
-          'The maximum size for on the fly area analysis is {number} km{sup}.',
+          'The maximum size for on the fly area analysis is {number} km.',
           {
             number: getLocaleNumber(HIGHER_AREA_SIZE_LIMIT, locale),
-            sup: <sup>2</sup>,
           }
         )}{' '}
-        {t('The area that you are trying to analyze has {number} km{sup}.', {
+        {t('The area that you are trying to analyze has {number} km.', {
```
Bluesmile82, line 60: I don't understand why are we removing all the ² (squares) on so many places. The units are still area so its km². Is there a reason?

src/utils/layer-manager-utils.js
```diff
@@ -209,16 +209,16 @@ export const createLayer = (layerConfig) => {
...
-    newLayer = new FeatureLayer({
-      portalItem: {
-        id: portalId
-      },
-      title: slug,
-      id: slug,
-      outFields: ['*'],
-    })
-  }else {
+  // if(portalId){
```
Bluesmile82, line 212: It would be better to remove this code if we are not going to use it

src/containers/sidebars/nrc-landing-sidebar/component.jsx
```diff
@@ -51,7 +51,7 @@ function NRCLandingSidebar({
       id: MARINE_SPI_FEATURE_LAYER,
       title: MARINE_SPI_FEATURE_LAYER,
       slug: MARINE_SPI_FEATURE_LAYER, // TODO: change specific slug for metadata
-      portalId: 'f5684ad998bf480aa1339763ff11670a',
+      portalId: '3f8accd89fe7482384972929fd2aae3f',
```
Bluesmile82, line 54: Are we using this portalIds at all, if we don't use them in layer manager utils?

Bluesmile82 (commented):
I think is better to merge to develop first and then to main to respect the git flow

src/containers/sidebars/data-global-sidebar/analyze-areas-sidebar-card/index.js
cmanciero: Was a request from Alex and Tamara to remove them

src/utils/layer-manager-utils.js
cmanciero: This code will come back once data is fixed. Arcgis seems to have a tough time displaying the color and styles of feature layers unless we used a portal id to load the layer.

cmanciero:
@Bluesmile82 Develop branch merged and then this branch was updated to follow the git flow

kkalkidan (approved):

PR #734, opened 2025-06-03
Fix borked code

cmanciero:
get fix done

### DIFF
--- a/src/components/country-entry-tooltip/country-entry-tooltip-component.jsx
+++ b/src/components/country-entry-tooltip/country-entry-tooltip-component.jsx
@@ -30,6 +30,8 @@ function CountryEntryTooltipComponent({
   onboardingType,
   waitingInteraction,
   isMobile,
+  activeTab,
+  setActiveTab,
 }) {
   const t = useT();
   const locale = useLocale();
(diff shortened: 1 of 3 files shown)


src/components/species-table/expanded-info/expanded-info.js
```diff
@@ -9,7 +9,7 @@ function ExpandedInfoContainer(props) {
   const { speciesName } = props;
   if (!speciesName) return null;
 
-  const url = `${VITE_APP_MOL_API}/species/info?scientificname=${speciesName}`;
+  const url = `https://api.mol.org/2.x/species/info?scientificname=${speciesName}`;
```
Bluesmile82, line 12: Why not use the ENV variable?

src/containers/nrc-content/nrc-content-component.jsx
```diff
@@ -239,7 +239,7 @@ function NrcContent({
 
               {!isMobile && (
                 <div className={styles.sourceText}>
-                  <p>{t('Source:')}</p>{' '}
+                  <p>{t('Source:')} Map of Life (2025)</p>{' '}
```
Bluesmile82, line 242: This could be added to Contentful as we have the sources defined there

PR #743, opened 2025-08-26
Fix race condition with labels

Bluesmile82:
### Description
This code waits for the label layers to be ready to style to them
### Testing instructions
To reproduce:

Go to localhost:3000/explore data
Open biodiversity
Click marine species global layers
Click richness tab
Click back to Priority then refresh the page
(No options are selected under the Priority tab)

When clicking on All Taxa under Global Layers, the layer should show

### DIFF
--- a/src/containers/layers/country-labels-layer/country-labels-layer-component.jsx
+++ b/src/containers/layers/country-labels-layer/country-labels-layer-component.jsx
@@ -37,21 +37,28 @@ function CountryLabelsLayerComponent(props) {
   }, [countryISO]);
 
   useEffect(() => {
-    const styleLayer = (layer) => {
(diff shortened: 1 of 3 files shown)


cmanciero (commented):
Screenshot 2025-08-26 at 1 09 53 PM
Checked out the branch and refreshed the page a couple times and seem to still be getting the error message.

Steps to reproduce:
1. Access Data globe
2. open Dev console
3. refresh page 
Once page finishes loading, if no error refresh again. Continue multiple times and the error will appear.

> 

1.

Bluesmile82:
Does the problem described still happen to you? Are the layers still not showing? I can see them every time

